PR #9, opened 2020-10-03
Build command

coffee-cup:
Lots of it taken from Suryas PR.

`railway build` command that dumps the env to .railway/env.json if the RAILWAY_PRODUCTION_TOKEN environment variable exists. We are not actually using the variable yet as a gql query needs to be built to handle it.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -18,7 +18,7 @@ func (h *Handler) Build(ctx context.Context, req *entity.CommandRequest) error {
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/build.go
```diff
@@ -0,0 +1,24 @@
...
+		fmt.Println("Railway env file is only generated in production")
+		return nil
+	}
+
+	err := h.ctrl.SaveEnvsToFile(ctx)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(fmt.Sprintf("Env written to %s \n Do NOT commit the env.json file. This command should only be run as a production build step.", h.cfg.RailwayEnvFilePath))
```
JakeCooper, line 21: Use multiline string with backticks here 

https://golang.org/ref/spec#String_literals

JakeCooper (approved):

PR #12, opened 2020-10-04
Remove build command, rename project token, and handle railway run with no command

coffee-cup:
- Remove build command as it is no longer needed
- Rename RAILWAY_PRODUCTION_TOKEN to RAILWAY_TOKEN
- Handle calling railway run with no command


errors/main.go
```diff
@@ -10,4 +10,5 @@ var (
 	ProjectNotFound       RailwayError = errors.New("Project not found.")
 	ProjectCreateFailed   RailwayError = errors.New("There was a problem creating the project")
 	ProductionTokenNotSet RailwayError = errors.New("RAILWAY_PRODUCTION_TOKEN environment variable not set")
```
JakeCooper, line 12: Change here as well

JakeCooper (approved):

PR #15, opened 2020-10-05
Support railway init projectid

coffee-cup:
- Support railway init <projectID>
- Prompt for environment if there is more than 1 in a project
- Show "Project not found" error instead of "graphql: Not authorised" if fetching a project fails

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -9,6 +9,27 @@ import (
 	"github.com/railwayapp/cli/ui"
 )
 
+func (h *Handler) saveProjectAndEnvironment(ctx context.Context, project *entity.Project) error {
+	if len(project.Environments) > 1 {
+		environment, err := ui.PromptEnvironments(project.Environments)
+		if err != nil {
+			return err
+		}
+
+		err = h.cfg.SetEnvironment(environment.Id)
+		if err != nil {
+			return err
+		}
+	} else if len(project.Environments) == 1 {
+		err := h.cfg.SetEnvironment(project.Environments[0].Id)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error {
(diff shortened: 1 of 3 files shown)


cmd/init.go
```diff
@@ -51,53 +70,63 @@ func (h *Handler) initFromAccount(ctx context.Context, req *entity.CommandReques
...
-		Environment: project.Environments[0].Id,
-	})
+	err = h.cfg.SetProject(project.Id)
+	if err != nil {
+		return err
+	}
 
+	err = h.saveProjectAndEnvironment(ctx, project)
 	if err != nil {
-		return nil
```
JakeCooper, line 61: Nice fix lol

cmd/init.go
```diff
@@ -51,53 +70,63 @@ func (h *Handler) initFromAccount(ctx context.Context, req *entity.CommandReques
...
+	}
+
+	return h.saveProjectWithID(ctx, projectID)
+}
+
 func (h *Handler) Init(ctx context.Context, req *entity.CommandRequest) error {
+	if len(req.Args) > 0 {
+		// projectID provided as argument
+		projectID := req.Args[0]
+		return h.saveProjectWithID(ctx, projectID)
```
JakeCooper, line 127: Can we keep these to following the common interface of taking a context + a CommandRequest?

projectId should be retrievable from req.Args[0]

gateway/project.go
```diff
@@ -39,7 +39,7 @@ func (g *Gateway) GetProject(ctx context.Context, projectId string) (*entity.Pro
 		Project *entity.Project `json:"projectById"`
 	}
 	if err := g.gqlClient.Run(ctx, gqlReq, &resp); err != nil {
-		return nil, err
+		return nil, errors.ProjectNotFound
```
JakeCooper, line 42: Any reason we're masking the error from users?

Additionally, it would be really good to have error reporting in the CLI

PR #20, opened 2020-10-05
Remove Home Entirely

JakeCooper:
* Until we have gco, user.Current is simply going to panic. I'm not sure why recover isn't working in that function, but this should be foolproof


coffee-cup (approved):
This is really weird. Not sure why I can't repro

PR #24, opened 2020-10-09
Templates

JakeCooper:
Adds the "template" command to the CLI. This fetches all the folders from the repository, allows you to visually scan through them, and auto clones and cleans up for you.

https://www.loom.com/share/3605630a33da4809bf9a28779057aadb

Originally added the v4 GH API, but the graphql stuff doesn't work unauthenticated so I've also added a v3 gateway for now.

### DIFF
--- a/cmd/template.go
+++ b/cmd/template.go
@@ -1,0 +1,49 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"os/exec"
+
+	"github.com/railwayapp/cli/entity"
+	"github.com/railwayapp/cli/ui"
+)
+
+func (h *Handler) Template(ctx context.Context, req *entity.CommandRequest) error {
+	dirs := []string{}
+	prompts := []string{"Select Language", "Select Type", "Select Template"}
+	for i := 0; i < 2; i++ {
+		options, err := h.ctrl.GetRailwayTemplates(ctx, strings.Join(dirs[:], "/"))
+		if err != nil {
+			return err
+		}
+		clearedOptions := []*entity.GithubFile{}
+		for _, file := range options {
+			if file.Type == "file" {
+				continue
+			}
+			clearedOptions = append(clearedOptions, file)
+		}
+		file, err := ui.PromptFiles(prompts[i], clearedOptions)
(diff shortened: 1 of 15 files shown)


cmd/template.go
```diff
@@ -0,0 +1,49 @@
...
+			clearedOptions = append(clearedOptions, file)
+		}
+		file, err := ui.PromptFiles(prompts[i], clearedOptions)
+		if err != nil {
+			return err
+		}
+		dirs = append(dirs, file.Name)
+	}
+	fmt.Println("Fetching Template...")
+	err := exec.Command("git", "clone", "git@github.com:railwayapp/examples.git", "delete-me").Run()
```
coffee-cup, line 36: In the goreleaser homebrew config, we should say that our package depends on git

cmd/template.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return err
+	}
+	fmt.Println("Generating Folder...")
+	err = exec.Command("mv", fmt.Sprintf("delete-me/%s", strings.Join(dirs[:], "/")), ".").Run()
+	if err != nil {
+		return err
+	}
+	fmt.Println("Cleaning Up")
+	err = exec.Command("rm", "-rf", "delete-me").Run()
```
coffee-cup, line 46: This won't work on Windows. Maybe use a stdlib call instead (same with above `mv` call)

ui/prompt.go
```diff
@@ -62,3 +62,45 @@ func PromptEnvironments(environments []*entity.Environment) (*entity.Environment
...
+
+	promptType := promptui.Select{
+		Label: "Select Language",
+		Items: []string{"Basic", "Blog", "Todo"},
+	}
+
+	_, _, err = promptType.Run()
+	if err != nil {
+		return "", err
+	}
```
coffee-cup, line 85: suggestion: remove these prompts if they are not used

cmd/template.go
JakeCooper: Good calls

JakeCooper:
Closing for now can re-open when we want to explore templating further

PR #25, opened 2020-10-15
Re-Enable up for Alpha

JakeCooper:
* Up enabled for Alpha Users﻿

### DIFF
--- a/controller/up.go
+++ b/controller/up.go
@@ -13,6 +13,10 @@ import (
 	"github.com/railwayapp/cli/entity"
 )
 
+var IGNORE_FILES = map[string]bool{
+	"node_modules": true,
+}
+
 func compress(src string, buf io.Writer) error {
 	// tar > gzip > buf
 	zr := gzip.NewWriter(buf)
@@ -35,7 +39,8 @@ func compress(src string, buf io.Writer) error {
 			return err
 		}
 		// if not a dir, write file content
-		if !fi.IsDir() {
+		_, ignore := IGNORE_FILES[fi.Name()]
+		if !fi.IsDir() && !ignore {
 			data, err := os.Open(file)
 			if err != nil {
 				return err
@@ -55,7 +60,6 @@ func compress(src string, buf io.Writer) error {
 	if err := zr.Close(); err != nil {
 		return err
 	}
-	//
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


gateway/up.go
```diff
@@ -0,0 +1,60 @@
...
+	client := &http.Client{}
+	resp, err := client.Do(httpReq)
+	if err != nil {
+		return nil, err
+	}
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode < 200 || resp.StatusCode > 400 {
```
coffee-cup, line 52:
```suggestion
	if resp.StatusCode < 200 || resp.StatusCode > 400 {
```
```suggestion
	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
```

controller/up.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	// produce gzip
+	if err := zr.Close(); err != nil {
+		return err
+	}
+	//
+	return nil
+}
+
+func (c *Controller) Up(ctx context.Context) error {
```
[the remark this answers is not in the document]
JakeCooper: That would mean exposing an endpoint to determine if a user is or isn't an admin. Too much leakage IMO.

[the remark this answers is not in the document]
coffee-cup: We already have a gql endpoint setup to fetch the user. We could a also return the `isAdmin` flag there. Alternatively, we could just check if there is any user config defined (admins will obviously be logged in)

controller/up.go
```diff
@@ -0,0 +1,89 @@
...
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/railwayapp/cli/entity"
+)
+
+var IGNORE_FILES = map[string]bool{
```
coffee-cup, line 16: I think this is good for now, but I still think that in the future we should use something like https://github.com/monochromegane/go-gitignore. Specifically to handle similar dep folders for other languages (or secret .env files).

coffee-cup (approved):

controller/up.go
JakeCooper: Done

PR #26, opened 2020-10-16
Versioning

sdan:
Adds versioning 

Pulls from ldflags and checks against github to see if there are any new tags


cmd/version.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (h *Handler) Version(ctx context.Context, req *entity.CommandRequest) error {
+	fmt.Println(fmt.Sprintf("railway version %s", constants.Version))
+	if constants.Version != "source" {
+		latest, err := getLatestVersion()
+		if err != nil {
+			return err
+		}
+		if latest != "" && latest != constants.Version {
+			fmt.Println("A newer version of the Railway CLI is available, please update to:", latest)
```
JakeCooper, line 20: Probs use printf with new line?

constants/version.go
```diff
@@ -0,0 +1,3 @@
+package constants
+
+var Version string = "source"
```
JakeCooper, line 3: Does this get overwritten, or is it source? IMO, if it’s piped in via LDflags, best to have an “empty” var with a comment of “piped in via LDflags on build”

JakeCooper (approved):
Provided it works and the comments are resolved this looks good

go.mod
```diff
@@ -4,13 +4,28 @@ go 1.14
...
+	github.com/spf13/cast v1.3.1 // indirect
 	github.com/spf13/cobra v1.0.0
+	github.com/spf13/jwalterweatherman v1.1.0 // indirect
+	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/spf13/viper v1.7.1
+	github.com/stretchr/testify v1.6.1 // indirect
+	golang.org/x/sys v0.0.0-20200915084602-288bc346aa39 // indirect
+	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
+	gopkg.in/ini.v1 v1.61.0 // indirect
+	gopkg.in/yaml.v2 v2.3.0 // indirect
```
JakeCooper, line 29: Lotta deps updates here. Anything in particular installed/got run?

main.go
```diff
@@ -43,59 +43,76 @@ func init() {
 	// Initializes all commands
 	handler := cmd.New()
 	rootCmd.AddCommand(&cobra.Command{
-		Use:   "login",
-		Short: "Login to Railway",
-		RunE:  contextualize(handler.Login),
+		Use:               "login",
+		Short:             "Login to Railway",
+		PersistentPreRunE: contextualize(handler.CheckVersion),
```
JakeCooper, line 48: Instead of doing this as a persistent pre run, can we not just do it on line 44 once? Also was gonna say check version doesn't need to be part of the handler but I think it does.

Sigh, we might need to refactor this CLI architecture to something more similar to what Stripe and GH are doing

constants/version.go
```diff
@@ -0,0 +1,3 @@
+package constants
+
+var Version string = "Piped into LDflags on build. You are probably running Railway CLI from source."
```
JakeCooper, line 3: Hmm, I like this but I also don't like the fact we have to do a string compare and copy it. Maybe something like
```suggestion
const VersionDefault = "Piped into LDflags on build. You are probably running Railway CLI from source."
var Version string = VersionDefault
```

Alternatively, we can make version a private variable and have a GetVersion function which returns the (version, error) and error when the version flag isn't set

Thoughts?

controller/version.go
```diff
@@ -0,0 +1,13 @@
+package controller
+
+import (
+	"context"
+
+	"github.com/google/go-github/github"
+)
+
+func (c *Controller) GetLatestVersion() (string, error) {
+	client := github.NewClient(nil)
```
JakeCooper, line 10: With the idiom we're using, client should be init'd inside the controller's New function

Whether that's the right idiom, for this, I'm growing to think not

go.mod
sdan: ran go mod tidy, not entirely sure why these pop up

JakeCooper:
Thoughts on having this just run for like, login and run? Seems like kinda spammy to do it everywhere

cmd/version.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/railwayapp/cli/entity"
+)
+
+func (h *Handler) Version(ctx context.Context, req *entity.CommandRequest) error {
+	fmt.Println(fmt.Sprintf("railway version %s", constants.Version))
+	return nil
+}
+
+func (h *Handler) CheckVersion(ctx context.Context, req *entity.CommandRequest) error {
+	if constants.Version != "Piped into LDflags on build. You are probably running Railway CLI from source." {
```
JakeCooper, line 17: Use VersionDefault here for comparison

JakeCooper (approved):
Approved pending only checking in `run` and `init`, as well as not doing string compare for version stuff

go.mod
JakeCooper: All good

PR #27, opened 2020-10-16
Visual improvements + clean up commands

sdan:
BRING THE COLORS

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -2,10 +2,21 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/railwayapp/cli/entity"
+	"github.com/railwayapp/cli/ui"
 )
 
 func (h *Handler) Up(ctx context.Context, req *entity.CommandRequest) error {
-	return h.ctrl.Up(ctx)
+	ui.StartSpinner(&ui.SpinnerCfg{
+		Message: "Accelerating your train to the clouds...",
+	})
+	url, err := h.ctrl.Up(ctx)
+	if err != nil {
+		return err
+	} else {
+		ui.StopSpinner(fmt.Sprintf("☁️ %s available at %s\n", ui.GreenText("Deploy"), ui.MagentaText(url)))
+	}
+	return nil
 }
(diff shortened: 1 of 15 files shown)


ui/text.go
```diff
@@ -15,3 +15,23 @@ func Bold(text string) string {
 	color := Color(os.Stdout)
 	return color.Sprintf(color.Bold(text))
 }
+
+func RedText(payload string) aurora.Value {
+	color := Color(os.Stdout)
+	return color.Red(payload).Bold()
```
coffee-cup, line 21: Should all the colours be bold as well? IMO the `RedText` function should only change the color.

ui/text.go
```diff
@@ -15,3 +15,23 @@ func Bold(text string) string {
 	color := Color(os.Stdout)
 	return color.Sprintf(color.Bold(text))
 }
+
+func RedText(payload string) aurora.Value {
+	color := Color(os.Stdout)
```
coffee-cup, line 20: Why is `os.Stdout` needed? Based on the aurora docs, would this work instead?

```go
func RedText(payload string) aurora.Value {
	return aurora.Red(payload)
}
```

coffee-cup (commented):
This is definitly an improvement. I do think that there is not enough consistency with the styles. For example,

image

In the above,
- the list select finish styles are different
- not sure why the text "Connected" is bold and purple

IMO (feel free to disagree), list select styles should finish like in "Connected to existing project" as we don't want to draw attention to them. Colors should not also be bold as well. And only a single section of text should be coloured per line as it should be used to draw the users attention to something specific.

JakeCooper:
Definitely fuck with the pink, but we should probs unify this

Before:
Screen Shot 2020-10-19 at 6 17 24 PM

After:
<green>Project:</green><pink>project.name</pink>
<green>Environment:

JakeCooper:
🤦

cmd/up.go
```diff
@@ -2,10 +2,21 @@ package cmd
...
+	"fmt"
 
 	"github.com/railwayapp/cli/entity"
+	"github.com/railwayapp/cli/ui"
 )
 
 func (h *Handler) Up(ctx context.Context, req *entity.CommandRequest) error {
-	return h.ctrl.Up(ctx)
+	ui.StartSpinner(&ui.SpinnerCfg{
+		Message: "Accelerating your train to the clouds...",
```
JakeCooper, line 13: Laying tracks in the clouds?

JakeCooper (approved):
Thumb

PR #29, opened 2020-10-16
Versioning fix

sdan:
Versioning not picking up`v` in the versions


sdan:
closing in favor of #30

PR #32, opened 2020-10-19
Add docs subcommand

sdan:
Adds docs subcommand under `open`
Also prints what commands are available when doing `railway open`


cmd/open.go
```diff
@@ -2,16 +2,27 @@ package cmd
...
 	"context"
+	"fmt"
 
 	"github.com/railwayapp/cli/entity"
 )
 
 func (h *Handler) Open(ctx context.Context, req *entity.CommandRequest) error {
-	projectId, err := h.cfg.GetProject()
+	if len(req.Args) == 0 {
+		fmt.Println("Use railway open to open links to Railway from the CLI. Here's whats we have:")
```
coffee-cup, line 12: This doesn't always work if the project can't be found

image

cmd/open.go
```diff
@@ -2,16 +2,27 @@ package cmd
...
+	}
+
+	projectCfg, err := h.cfg.GetProjectConfigs()
+	if err != nil {
+		return err
+	}
+	project, err := h.ctrl.GetProject(ctx, projectCfg.Project)
```
coffee-cup, line 19: You should be able to open the docs if the project is not found

image

controller/open.go
```diff
@@ -0,0 +1,73 @@
...
+	} else {
+		return errors.OpenCommandNotFound
+	}
+	return nil
+
+}
+
+func nameList(id string) error {
+	names, longest := getNames()
+	fmt.Printf("%s%s\n", padName("shortcut", longest), "    url")
```
coffee-cup, line 35: I'm not sure its is clear that the shortcuts should be used with railway open

Maybe 

instead of just "docs" and "project" it is "railway open docs" and "railway open project"?

JakeCooper:
I think, to Jake's comment, this would be better off as a secondary docs command with railway open just working for project

Thoughts?

sdan:
closing in favor of opening up a docs command

PR #33, opened 2020-10-19
Browserless login

coffee-cup:
Ability to login to Railway when the CLI cannot open a browser directly

See https://github.com/railwayapp/mono/pull/253 for an explanation of how this works

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/railwayapp/cli/entity"
+	"github.com/railwayapp/cli/ui"
 )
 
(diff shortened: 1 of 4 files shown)


JakeCooper (approved):
Looks good. Only thing I'd recommend is, to include a message on the browser based loading screen like "Browser didn't open? Try rerunning with --browserless"

IDK up to you. I was stoked about the idea but midway through typing it I was like "ehh"

coffee-cup:
I think if the browser doesn't open it will error and we will go through the browserless flow anyway. I'll probably not add that message just because it feels less magical ✨

PR #34, opened 2020-10-20
URL consent

sdan:
Ask for user to hit enter before going to URL. Will update when #32 is merged to include open.

### DIFF
--- a/controller/user.go
+++ b/controller/user.go
@@ -96,10 +96,7 @@ func (c *Controller) Login(ctx context.Context) (*entity.User, error) {
 	}()
 	url := getLoginURL(port, code)
 
-	err = confirmBrowserOpen("Logging in...", url)
-	if err != nil {
-		return nil, err
-	}
+	confirmBrowserOpen("Logging in...", url)
 
 	wg.Wait()
 	err = c.cfg.SetUserConfigs(&entity.UserConfig{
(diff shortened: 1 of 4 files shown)


controller/user.go
```diff
@@ -134,6 +141,22 @@ func (c *Controller) IsLoggedIn(ctx context.Context) (bool, error) {
 	return isLoggedIn, nil
 }
 
+func confirmBrowserOpen(spinnerMsg string, url string) error {
```
JakeCooper, line 144: Function never technically returns an error right? Should just make it return nothing and save the err != nil check

JakeCooper (approved):

controller/user.go
JakeCooper, line 144:
```suggestion
func confirmBrowserOpen(spinnerMsg string, url string) {
```

controller/user.go
```diff
@@ -134,6 +141,22 @@ func (c *Controller) IsLoggedIn(ctx context.Context) (bool, error) {
...
+		Message: spinnerMsg,
+	})
+	err := browser.OpenURL(url)
+	if err != nil {
+		ui.StopSpinner(fmt.Sprintf("Failed to open browser, please go to %s manually.", url))
+		ui.StartSpinner(&ui.SpinnerCfg{
+			Message: spinnerMsg,
+		})
+	}
+	return nil
```
JakeCooper, line 157:
```suggestion
```

PR #43, opened 2020-10-22
Open to environment

JakeCooper:
* When opening, pass the environment as a query strong so that it always sets your environment correctly﻿


sdan (approved):
👍

PR #47, opened 2020-10-29
Use star for origin headers

JakeCooper:


JakeCooper:
kk @coffee-cup with the precog PR

PR #53, opened 2020-11-19
Move cli configs into root dir

coffee-cup:
- Migrate local project ./railway config to user level ~/.railway config
- When looking for project config, find closest parent directory that has been init'd with a railway project
- When initing a new railway project, always use current working directory

The top level config has been split into user config and project config. The APIs for reading/writing user/project configs have stayed the same. But their implementations now call into the root config which deals with marsheling/unmarshelling the configs

The top level config will now look something like

```json
{
  "user": {
    "token": ""
  },
  "projects": {
    "/users/jakerunzer/dev/railway-cli": {
      "projectPath": "/users/jakerunzer/dev/railway-cli",
      "project": "XXX",
      "environment": "XXX"
    },
    "/users/jakerunzer/dev/railway-cli/node_modules": {
      "projectPath": "/users/jakerunzer/dev/railway-cli/node_modules",
      "project": "XXX",
      "environment": "XXX"
    }
  }
}
```

image

### DIFF
--- a/configs/main.go
+++ b/configs/main.go
@@ -1,11 +1,10 @@
 package configs
 
 import (
-	"encoding/json"
-	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
+	"reflect"
 
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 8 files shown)


configs/main.go
```diff
@@ -48,24 +49,46 @@ func (c *Configs) unmarshalConfig(config *Config, data interface{}) error {
...
-		config.viper.Set(k, v)
-	}
+	// reflectCfg := reflect.ValueOf(cfg)
+	// for i := 0; i < reflectCfg.NumField(); i++ {
+	// 	k := reflectCfg.Type().Field(i).Name
+	// 	v := reflectCfg.Field(i).Interface()
+	// 	config.viper.Set(k, v)
+	// }
+
+	// err = config.viper.WriteConfig()
```
JakeCooper, line 59: Remove if not in use
```suggestion
```

configs/main.go
```diff
@@ -48,24 +49,46 @@ func (c *Configs) unmarshalConfig(config *Config, data interface{}) error {
...
-	err = config.viper.WriteConfig()
+	jsonString, err := json.MarshalIndent(cfg, "", "  ")
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(config.configPath, jsonString, os.ModePerm)
+	if err != nil {
+		return err
+	}
```
JakeCooper, line 74: If we're using viper still, shouldn't we be doing this via viper? I'm not a huge fan of having two mechanisms to do io

configs/project.go
```diff
@@ -1,55 +1,189 @@
...
+			pathMatch = path
+			break
+		}
+
+		// check if parent path matches
+		var matches = strings.HasPrefix(cwd, fmt.Sprintf("%s/", path))
+		if matches && len(path) > longestPath {
+			longestPath = len(path)
+			pathMatch = path
+		}
```
JakeCooper, line 87: These could probably be combined no? Simply use string contains, or maybe even prefix will handle it

JakeCooper (approved):
Looks good other than a few small things

Ideally, we should either use Viper, or be very express about the safety hatch we're using (shove it in it's own function with a comment about how Viper can't really do nested configs). It would really surprise me if it couldn't tho...

configs/project.go
coffee-cup: The reason I separated it is because of the trailing slash. The current directory does not have a trailing slash but we want to make sure that `/this/path` does not match `/this/pathalso`.

Thinking of it now I can just add a trailing slash to the current directory path and combine the two checks :)

configs/main.go
coffee-cup: Yeah I agree it isn't ideal to go around viper and just write the json manually. But I couldn't figure out how to just write an entire config file with viper and doing the recursive reflect stuff seems more complicated than it needs to be. I'll do some digging in the viper docs and try and find a better solution.

JakeCooper (approved):

PR #56, opened 2020-11-24
Use width of longest env var name when printing

coffee-cup:
image

I know we don't want a utils folder, but min/max for ints should probably live somewhere else?


JakeCooper (approved):
Probs fine for now, at least until we need to move it somewhere

Big fan of WET (Write Everything Twice)

PR #58, opened 2020-12-01
Install and uninstall shell scripts

coffee-cup:
Shell script that can be installed via curl and sh. The user can use it like

```shell
sh -c "$(curl -sSL https://github.com/railwayapp/cli/blob/master/install.sh)"
```

The latest version will be used by default, but you can specify the `VERSION` environment variable to install a specific version of railway to install.

The install process is

- Create tmp directory
- Download latest asset from GH to tmp
- Compare checksums
- Extract asset
- Install to /usr/local/bin
- Remove tmp

There is also an uninstall script that remove `/usr/local/bin/railway`

image

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -37,7 +37,7 @@ get_checksum_url() {
 }
 
 command_exists() {
-	command -v "$@" >/dev/null 2>&1
+  command -v "$@" >/dev/null 2>&1
 }
 
 fmt_error() {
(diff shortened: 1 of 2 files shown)


coffee-cup:
Tested via curl to this branch and it works

```
sh -c "$(curl -sSL https://raw.githubusercontent.com/railwayapp/cli/j/rail-484-install-cli-from-pipe-command/install.sh)"
```

image

JakeCooper (approved):

PR #60, opened 2020-12-04
Guard against NPE on version check

JakeCooper:
If the version check code errors, we should bubble up that error before we deref

### DIFF
--- a/controller/version.go
+++ b/controller/version.go
@@ -9,5 +9,5 @@ func (c *Controller) GetLatestVersion() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return *rep.TagName, err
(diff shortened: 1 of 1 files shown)


coffee-cup (approved):

controller/version.go
```diff
@@ -6,5 +6,8 @@ import (
 
 func (c *Controller) GetLatestVersion() (string, error) {
 	rep, _, err := c.ghc.Repositories.GetLatestRelease(context.Background(), "railwayapp", "cli")
+	if err != nil {
+		return "", err
+	}
 	return *rep.TagName, err
```
JakeCooper, line 12:
```suggestion
	return *rep.TagName, nil
```

PR #65, opened 2021-02-02
fix: Don't delete .railway/ if running in ~/

queer:


configs/project.go
```diff
@@ -20,6 +20,11 @@ func (c *Configs) MigrateLocalProjectConfig() error {
 		return err
 	}
 
+	// Avoid deleting ~/.railway
+	if projectDir == os.Getenv("HOME") {
```
coffee-cup, line 24: I think this might be an issue on Windows computers? It might be worth checking if `HOME` is empty too. We have issues with Windows support in the past -> https://railway.canny.io/admin/feedback/feature-requests/p/powershell-support?boards=feature-requests

coffee-cup (approved):

configs/project.go
queer: Ah hm, good point. That being said, I don't have a Windows box to test on :pensive:

PR #68, opened 2021-02-20
Allow installation with custom binary name and install location

nickpwhite:
I'm running Ubuntu 20.04 and can't install via curl without `sudo` since `/usr/local/bin` is owned by root. These changes allow the user to specify an install directory (and binary name) optionally with environment variables. I also included a change to make the checksum test more compatible across different sh implementations. Bash/zsh implement `[[` but it's not in the sh spec. `[` is equivalent here.


JakeCooper (approved):
😍

PR #69, opened 2021-03-04
feature: Build and run Dockerfile if no args given to `railway run`

queer:


cmd/run.go
```diff
@@ -40,3 +55,87 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
...
+
+	project, err := h.ctrl.GetProject(ctx, projectCfg.Project)
+	if err != nil {
+		return err
+	}
+
+	// Strip characters not allowed in Docker image names
+	sanitiser := regexp.MustCompile(`[^A-Za-z0-9_-]`)
+	imageName := sanitiser.ReplaceAllString(project.Name, "")
+	image := fmt.Sprintf("railway-local/%s:latest", imageName)
```
JakeCooper, line 74: Any reason for Railway-local vs just Railway? Our internal representation for image routes is Railway/ so I could see it either way (nice to differentiate but also maybe nice to unify?)

queer: Mostly just to differentiate. imo it makes it a bit clearer that this isn't the image that's actually running up there in Railway, but it's very specifically one built locally.

cmd/run.go
```diff
@@ -40,3 +55,87 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
...
+
+	out, err := buildCmd.CombinedOutput()
+	if err != nil {
+		ui.StopSpinner("")
+		return showCmdError(buildCmd.Args, out, err)
+	}
+
+	ui.StopSpinner(fmt.Sprintf("🎉 Built %s", image))
+
+	port := 4411
```
JakeCooper, line 97: Rng port vs 3000 vs 4411? Can see all of em with pros and cons

queer: I'd probably go RNG if I had to pick again; 4411 was picked because it gets pretty close to looking like "RAIL"

cmd/run.go
```diff
@@ -40,3 +55,87 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
...
+	runCmd.Stdout = os.Stdout
+	runCmd.Stderr = os.Stdout
+	runCmd.Stdin = os.Stdin
+
+	err = runCmd.Run()
+	if err != nil {
+		return err
+	}
+
+	// TODO: Probably should be cleaning up the image here...
```
JakeCooper, line 120: +1

JakeCooper (approved):

cmd/run.go
JakeCooper: Yup, I can totally vibe with this

PR #71, opened 2021-03-09
Add connect commands + selector

queer:
- Builds commands for each plugin from project env
- Provides menu for selecting plugin to connect to

### DIFF
--- a/cmd/connect.go
+++ b/cmd/connect.go
@@ -42,15 +42,19 @@ func (h *Handler) Connect(ctx context.Context, req *entity.CommandRequest) error
 		plugin = req.Args[0]
 	}
 
+	if !isPluginValid(plugin) {
+		return fmt.Errorf("Invalid plugin: %s", plugin)
+	}
 	envs, err := h.ctrl.GetEnvs(ctx)
 	if err != nil {
 		return err
 	}
-	if !isPluginValid(plugin) {
-		return fmt.Errorf("Invalid plugin: %s", plugin)
-	}
 
 	command, connectEnv := buildConnectCommand(plugin, envs)
+	if !commandExistsInPath(command[0]) {
+		fmt.Println("🚨", ui.RedText(command[0]), "was not found in $PATH.")
+		return nil
+	}
 
 	cmd := exec.Command(command[0], command[1:]...)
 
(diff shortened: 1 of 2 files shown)


cmd/connect.go
```diff
@@ -0,0 +1,114 @@
...
+
+	fmt.Printf("🎉 Connecting to: %s %s\n", ui.MagentaText(project.Name), ui.MagentaText(envName))
+
+	names := make([]string, 0)
+	for _, plugin := range project.Plugins {
+		// TODO: Better way of handling this
+		if plugin.Name != "env" {
+			names = append(names, plugin.Name)
+		}
+	}
```
queer, line 31: This is a bit of an ugly thing here, but I just didn't see a better way to handle this...

JakeCooper: @coffee-cup Another env oopsie :<

Wonder if there's a way to move this before it becomes a "founding principle"

JakeCooper: e.g we ship binaries with logic for it

coffee-cup: We should probably have a `visiblePlugins` function that filters out env plugins. I can see at one point we would have multiple env plugins.

JakeCooper (approved):
I think this is dope, but I would also want a "hot" command that I could just type to connect. Thoughts?

`railway psql` or `railway connect psql` or `railway run psql`

I think the latter COULD be good, because it would seem as thought we're truly running that command "in the environment". My only thoughts is kinda overloading the "run" command

cmd/connect.go
coffee-cup: could also just stop sending back "env" as part of the project plugins. But I imagine that would break a few other things

coffee-cup:
Should Wondering if we should have some sort of help message if the connect command fails? Something like "Connecting failed? Do you have psql installed?"

cmd/connect.go
coffee-cup: Actually, in this case we might want a `getDatabasePlugins` since we will (hopefully) have other non-db plugins in the future

JakeCooper:
Yea maybe a little error message. Obvs stdout will have it but still might be nice.

queer:
> I think this is dope, but I would also want a "hot" command that I could just type to connect. Thoughts?

:+1: Added.

PR #73, opened 2021-03-19
Client-side Protected Environments

JakeCooper:
Currently, no way to "unprotect" an environment other than editing config.json

### DIFF
--- a/cmd/protect.go
+++ b/cmd/protect.go
@@ -1,0 +1,39 @@
+package cmd
+
+import (
+	"context"
+
+	"github.com/railwayapp/cli/entity"
+	"github.com/railwayapp/cli/ui"
+)
+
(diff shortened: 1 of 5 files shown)


cmd/protect.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		return err
+	}
+
+	project, err := h.ctrl.GetProject(ctx, projectID)
+	if err != nil {
+		return err
+	}
+
+	environment, err := ui.PromptEnvironments(project.Environments)
```
coffee-cup, line 26: Hmm, I would expect this command to just protect the current environment. Thoughts?

coffee-cup (commented):
I feel like we should also have an unprotect command. Since its an undocumented feature it is fine for now. But I am not a fan of suggesting users to edit the config.json

coffee-cup (approved):

PR #75, opened 2021-03-24
Deploy from Starter Template

gschier:
This PR adds the ability to `init` a project from one of the [Starter Templates](https://railway.app/starters).

It will:

1. Fetch starter templates data
2. Prompt user to select starter
3. Prompt user to select GitHub org
4. Prompt user for project name
5. Select public/private
5. Prompt user for env vars (if needed)
6. Create project and wait for workflow to finish
7. Open project in default browser

image

image

image

image

image

image


cmd/init.go
```diff
@@ -43,6 +44,112 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
...
+	// Prompt for a template to use
+
+	template, err := ui.PromptStarterTemplates()
+	if err != nil {
+		return err
+	}
+
+	// Select GitHub org
+
+	ui.StartSpinner(&ui.SpinnerCfg{Message: "Fetching GitHub scopes"})
```
JakeCooper, line 57: ultranit: newline for readability/simple extension
```suggestion
	ui.StartSpinner(&ui.SpinnerCfg{
	    Message: "Fetching GitHub scopes",
	})
```

cmd/init.go
```diff
@@ -4,13 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
 	"github.com/railwayapp/cli/entity"
+	errors2 "github.com/railwayapp/cli/errors"
```
JakeCooper, line 8: Thoughts on `CliErrors` or `RailwayErrors`?

cmd/init.go
```diff
@@ -43,6 +44,112 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
...
+	ui.StopSpinner("")
+
+	org, err := ui.PromptGitHubScopes(scopes)
+	if err != nil {
+		return err
+	}
+
+	// Enter project name
+
+	name, err := ui.PromptProjectName()
```
JakeCooper, line 74: handle err here (Damn it Golang why don't you warn us about these things!)

Maybe it's time we consider adding a [Linter](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/errcheck.go)

cmd/init.go
```diff
@@ -43,6 +44,112 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
...
+				return err
+			}
+			variables[name] = v
+		}
+		fmt.Print("\n")
+	}
+
+	// Create Railway project
+
+	ui.StartSpinner(&ui.SpinnerCfg{Message: "Creating project"})
```
JakeCooper, line 94: Same linesplit nit. Your call obvs
```suggestion
	ui.StartSpinner(&ui.SpinnerCfg{
	    Message: "Creating project",
	})
```

cmd/init.go
```diff
@@ -43,6 +44,112 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
...
+		time.Sleep(2 * time.Second)
+		ws, err := h.ctrl.GetWorkflowStatus(ctx, creationResult.WorkflowID)
+		if err != nil {
+			return err
+		}
+		if ws.Status == "Error" {
+			ui.StopSpinner("")
+			return errors2.WorkflowFailed
+		}
+		if ws.Status == "Complete" {
```
JakeCooper, line 129: I'm unsure if we use these anywhere else, but we might want to hoist them into an enum and match against that vs doing direct string comp

cmd/init.go
```diff
@@ -43,6 +44,112 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
...
+	})
+
+	for {
+		time.Sleep(2 * time.Second)
+		ws, err := h.ctrl.GetWorkflowStatus(ctx, creationResult.WorkflowID)
+		if err != nil {
+			return err
+		}
+		if ws.Status == "Error" {
+			ui.StopSpinner("")
```
JakeCooper, line 126:
```suggestion
			ui.StopSpinner(fmt.Sprintf("Uhh Ohh. Workflow failed!\nError:", err.Error()))
```

gateway/project.go
```diff
@@ -71,6 +71,34 @@ func (g *Gateway) CreateProject(ctx context.Context, req *entity.CreateProjectRe
...
+	gqlReq := gql.NewRequest(`
+		mutation($name: String!, $org: String!, $template: String!, $isPrivate: Boolean, $plugins: [String!], $variables: Json) {
+			createProjectFromTemplate(name: $name, org: $org, template: $template, isPrivate: $isPrivate, plugins: $plugins, variables: $variables) {
+				projectId
+				workflowId
+			}
+		}
+	`)
+
+	g.authorize(ctx, gqlReq.Header)
```
JakeCooper, line 84: I think this bad boi returns an err?
```suggestion
	err := g.authorize(ctx, gqlReq.Header)
	if err != nil {
	    return nil, err
	}
```

gateway/scope.go
```diff
@@ -0,0 +1,30 @@
...
+
+// GetWritableGithubScopes returns scopes associated with Railway user
+func (g *Gateway) GetWritableGithubScopes(ctx context.Context) ([]string, error) {
+	gqlReq := gql.NewRequest(`
+		query {
+			getWritableGithubScopes 
+		}
+	`)
+
+	err := g.authorize(ctx, gqlReq.Header)
```
JakeCooper, line 17: yeeee, seems to

gateway/workflow.go
```diff
@@ -0,0 +1,30 @@
...
+func (g *Gateway) GetWorkflowStatus(ctx context.Context, workflowID string) (*entity.WorkflowStatus, error) {
+	gqlReq := gql.NewRequest(`
+		query($workflowId: String!) {
+			getWorkflowStatus(workflowId: $workflowId) {
+				status
+			}
+		}
+	`)
+
+	g.authorize(ctx, gqlReq.Header)
```
JakeCooper, line 19:
```suggestion
	err := g.authorize(ctx, gqlReq.Header)
	if err != nil {
	    return nil, err
	}
```

ui/prompt.go
```diff
@@ -44,18 +48,91 @@ func PromptProjects(projects []*entity.Project) (*entity.Project, error) {
 		Templates: &promptui.SelectTemplates{
 			Active:   `{{ .Name | underline }}`,
 			Inactive: `{{ .Name }}`,
-			Selected: fmt.Sprintf("%s Project: {{ .Name | magenta | bold }} ", GreenText("✔")),
+			Selected: fmt.Sprintf("%s Project: {{ .Name | magenta | bold }} ", promptui.IconGood),
```
JakeCooper, line 51: Nice

ui/prompt.go
```diff
@@ -44,18 +48,91 @@ func PromptProjects(projects []*entity.Project) (*entity.Project, error) {
...
 		},
 	}
 	i, _, err := prompt.Run()
 	return projects[i], err
 }
 
+// PromptStarterTemplates fetches available templates and prompts the user to select one
+func PromptStarterTemplates() (*entity.Template, error) {
+	StartSpinner(&SpinnerCfg{Message: "Fetching starter templates"}) // Fetch starter templates
```
JakeCooper, line 60:
```suggestion
	StartSpinner(&SpinnerCfg{
	    Message: "Fetching starter templates",
	}) // Fetch starter templates
```

controller/scope.go
```diff
@@ -0,0 +1,10 @@
+package controller
+
+import (
+	"context"
+)
+
+// GetWriteableGithubScopes creates a project specified by the project request, error otherwise
```
JakeCooper, line 7:
```suggestion
// GetWritableGithubScopes creates a project specified by the project request, error otherwise
```

cmd/init.go
```diff
@@ -43,6 +44,112 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
...
+	}
+	if len(scopes) == 0 {
+		return errors2.NoGitHubScopesFound
+	}
+	ui.StopSpinner("")
+
+	org, err := ui.PromptGitHubScopes(scopes)
+	if err != nil {
+		return err
+	}
```
JakeCooper, line 70: Thoughts on, in the event of a single scope, just selecting it for the user? IMO, the less steps the better
```suggestion
        if len(scopes) == 1 {
            fmt.Printf("Using %s GitHub Account\n", scopes[0])
        } else {
            org, err := ui.PromptGitHubScopes(scopes)
	    if err != nil {
	        return err
	    }
        }
```

JakeCooper (approved):
I know that's a lot of comments but all of them are nits

Looks great :D

cmd/init.go
gschier: CLIErrors is good. IDE named it `errors2`

cmd/init.go
gschier: This logic is in `ui.PromptGitHubScopes`

cmd/init.go
gschier: Ugh, yes!

cmd/init.go
gschier: Don't have access to `err` here but will add a the log message 👍🏻

controller/scope.go
gschier: Damn, how'd  you spot that one?

JakeCooper: I actually searched for "WritableGithubScopes" and also misspelled it, and realized that there was only a comment for this

Total luck really haha

gateway/project.go
gschier: copy-pasta. Will update the others too

cmd/init.go
JakeCooper: Oh right that's the err for the request not the error as a response

controller/scope.go
gschier: 😆

cmd/init.go
JakeCooper: Oh perfecto

cmd/init.go
```diff
@@ -43,6 +44,124 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
...
+		for _, envVar := range template.EnvVars {
+			fmt.Printf("\n(%s)\n", ui.GrayText(envVar.Desc))
+			v, err := ui.PromptText(envVar.Name)
+			if err != nil {
+				return err
+			}
+			variables[name] = v
+		}
+		fmt.Print("\n")
+	}
```
coffee-cup, line 100: Environment variables can be optional and/or have a default value. This should probably handle that as well.
For optional env vars I think we could show a "(optional. enter to leave blank)" text (or something similar). For env vars with a default value, maybe we can auto-fill their input?

gschier: Oo, will add that, nice catch

entity/template.go
```diff
@@ -0,0 +1,15 @@
...
+	Text    string   `json:"text"`
+	Href    string   `json:"href"`
+	Plugins []string `json:"plugins"`
+	Desc    string   `json:"desc"`
+	Tags    []string `json:"tags"`
+	Icon    string   `json:"icon"`
+	Value   string   `json:"value"`
+	EnvVars []struct {
+		Name string `json:"name"`
+		Desc string `json:"desc"`
```
coffee-cup, line 13: Similar to above comment, can have optional and default here. The type we use on the frontend is

image

cmd/init.go
```diff
@@ -43,6 +44,124 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
...
+		return err
+	}
+
+	err = h.cfg.SetEnvironment(environment.Id)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("🎉 Created project %s\n", ui.MagentaText(name))
+	h.ctrl.OpenProjectInBrowser(ctx, project.Id, environment.Id)
```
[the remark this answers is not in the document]
gschier: Ah, nice! I was wondering why it opened to the Postgres plugin by default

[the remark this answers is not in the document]
coffee-cup: Ya the current logic when you just redirect to a project is to open the first plugin. We might want to change that to the deployments page as it seems we are focusing more on that (first versions of Railway were almost all plugins).

We have also been thinking about what a project overview page would look like. (obvi out of scope of this PR, just some context).

ui/prompt.go
```diff
@@ -44,18 +48,141 @@ func PromptProjects(projects []*entity.Project) (*entity.Project, error) {
...
+}
+
+func PromptEnvVars(envVars []entity.TemplateEnvVar) (map[string]string, error) {
+	variables := make(map[string]string)
+	if len(envVars) > 0 {
+		fmt.Printf("\n%s\n", Bold("Environment Variables"))
+	}
+
+	requiredValidator := func(s string) error {
+		if s == "" {
```
coffee-cup, line 109: suggestion: trim the value here. A bunch of whitespace should be invalid for a required env var

coffee-cup (approved):
Small suggestion, but besides that it looks good!

PR #77, opened 2021-03-30
Clean up global commands and help text

gschier:
Changes:

- Renamed `disconnect` to `unlink` and split `init` into `init` and `link`
- Renamed `env` to `variables` to align verbiage with Dashboard
- Adjusted help text for all commands
- Both Stripe and Heroku CLIs were used as a style reference

image

Old renamed command still work but are deprecated.

image

### DIFF
--- a/main.go
+++ b/main.go
@@ -52,7 +52,7 @@ func init() {
 
 	loginCmd := &cobra.Command{
 		Use:   "login",
-		Short: "Login to Railway",
+		Short: "Login to your Railway account",
 		RunE:  contextualize(handler.Login, handler.Panic),
 	}
 	loginCmd.Flags().Bool("browserless", false, "--browserless")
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -52,61 +52,73 @@ func init() {
...
-		Short: "Disconnect from Railway",
+		Use:        "disconnect",
+		Short:      "Disconnect from Railway",
+		RunE:       contextualize(handler.Disconnect, handler.Panic),
+		Deprecated: "The 'railway disconnect' command is now railway 'disassociate'", /**/
+	})
+	rootCmd.AddCommand(&cobra.Command{
+		Use:   "disassociate",
+		Short: "Remove association between project and current directory",
 		RunE:  contextualize(handler.Disconnect, handler.Panic),
```
coffee-cup, line 86: suggestion: rename the handler function to `Disassociate`

coffee-cup (approved):

JakeCooper:
I'm unsure how I feel about `init` and `disassociate` being complementary commands. They don't sound related and IDK how I feel about the verbage of `disassociate`

To me, it almost feels like they should be `connect` and `disconnect` and we could consider renaming `connect` to something else

Thoughts?

gschier:
@JakeCooper ya, it still seems a bit weird to me too. I was also thinking that `disassociate` may be an infrequent-enough command to put under `railway init remove` or something.

gschier:
We could rename `init` to `link/unlink` but that doesn't imply the ability to create a project as much as `init` does. Or we could rename `disassociate` to `unlink` (didn't think of that verb before).

PR #78, opened 2021-03-30
Revert config path handling and update DNE logic

gschier:
I had originally changed the config loading logic in #77 in order to have `ReadInConfig` return `viper.ConfigFileNotFoundError`. However, this broke things because `Config.configPath` would be unset if the project config didn't exist yet, breaking `railway run` (and possibly other commands).

This PR reverts back to the old way of loading configs with a fixed path and uses `os.IsNotExist()` to check the error instead

### DIFF
--- a/configs/main.go
+++ b/configs/main.go
@@ -80,7 +80,7 @@ func New() *Configs {
 	if os.IsNotExist(err) {
 		// That's okay, configs are created as needed
 	} else if err != nil {
-		fmt.Printf("Unable to parse railway config! %#v\n", err)
+		fmt.Printf("Unable to parse railway config! %s\n", err)
 	}
 
 	rootConfig := &Config{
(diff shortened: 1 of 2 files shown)


configs/main.go
```diff
@@ -94,21 +96,18 @@ func New() *Configs {
...
+		// That's okay, continue
 	} else if err != nil {
-		// Config file was found but another error was produced
-		fmt.Printf("Unable to load project config! %#v\n", err)
+		fmt.Println("Unable to load project config!")
 	}
 
 	projectConfig := &Config{
 		viper:      projectViper,
-		configPath: projectViper.ConfigFileUsed(),
```
gschier, line 111: This was the crux of the issue. If a config wasn't found, this path would be empty.

JakeCooper (approved):

configs/project.go
```diff
@@ -61,10 +61,10 @@ func (c *Configs) getCWD() (string, error) {
 }
 
 func (c *Configs) GetProjectConfigs() (*entity.ProjectConfig, error) {
-	err := c.MigrateLocalProjectConfig()
```
gschier, line 64: This was also an issue, added in #76

PR #80, opened 2021-03-30
Display team projects in link command

coffee-cup:
- Handle when there are no projects found
- remove avatar and projects from user entity
- display team projects in link command

It would be nice to eventually show which team the project is in when they are listed out. But for now, this PR just fixes the team projects not being selectable in the `railway link` project list.

### DIFF
--- a/cmd/link.go
+++ b/cmd/link.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"context"
+	"fmt"
+
 	"github.com/railwayapp/cli/entity"
 	"github.com/railwayapp/cli/ui"
 )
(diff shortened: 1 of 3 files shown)


JakeCooper:
> It would be nice to eventually show which team the project is in when they are listed out

This wouldn't be too bad given we know scope of the thing right? I think, if there are teams, we should iter through and slap a "Team Name" as a *string onto the entity on this side, which we can display in color if it exists (and only add it if there's  team & user projects)

I can add this if you don't want to as part of this PR

JakeCooper (approved):

PR #82, opened 2021-04-01
Add `variables` `get|set|delete` subcommands

gschier:
This PR adds the ability to modify variables from the CLI.

- New `variables  get|set|delete`
  - Behavior was borrowed from the Heroku CLI)
- Correctly handle protected environments
- Removed `getEnvironmentNameFromID` helper function (replaced with `Controller.GetEnvironment()`)
- Added a bunch of formatting helpers to `text.go` for things like headings, alerts, etc.
- Added hidden `design` command to print a formatting "styleguide" 

_NOTE: This PR really contains two features (design stuff and variables). It's not that large but please let me know if you'd prefer to break into two PRs._

`set` updates the environment and prints the values that were changed

image

`get` prints the values of the given variables names, each on a newline. (Doesn't print anything else for piping use-case)

image

`delete` deletes the variables and prints a confirmation message
image

`design` prints a bunch of common styles

image

### DIFF
--- a/cmd/design.go
+++ b/cmd/design.go
@@ -13,7 +13,7 @@ func (h *Handler) Design(ctx context.Context, req *entity.CommandRequest) error 
 	fmt.Print(ui.AlertWarning("That might not hve been what you wanted"))
 	fmt.Print(ui.AlertInfo("Just so you know you know, Railway is awesome"))
 
-	println("")
+	fmt.Println("")
 	fmt.Print(ui.Heading("Unordered List"))
 	fmt.Print(ui.UnorderedList([]string{
 		"List Item 1",
@@ -21,14 +21,14 @@ func (h *Handler) Design(ctx context.Context, req *entity.CommandRequest) error 
 		"List Item 3",
 	}))
 
-	println("")
+	fmt.Println("")
 	fmt.Print(ui.Heading("Ordered List"))
 	fmt.Print(ui.OrderedList([]string{
 		"First Step",
 		"Next Step",
 	}))
 
-	println("")
+	fmt.Println("")
 	fmt.Print(ui.Heading("Key-Value Pairs"))
 	fmt.Print(ui.KeyValues(map[string]string{
 		"First Key":  "Value 1",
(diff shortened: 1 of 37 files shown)


cmd/design.go
```diff
@@ -0,0 +1,66 @@
...
+
+	println("")
+	fmt.Print(ui.Heading("Indented"))
+	fmt.Print(ui.Indent("func main() {\n  println(\"Hello World!\")\n}"))
+
+	println("")
+	fmt.Print(ui.Heading("Block Quote"))
+	fmt.Print(ui.BlockQuote("That's the thing about counter-intuitive ideas. They contradict your intuitions. So, they seem wrong"))
+
+	println("")
```
JakeCooper, line 61: Does this work? I’ve always used fmt

controller/envs.go
```diff
@@ -29,10 +29,13 @@ func (c *Controller) GetEnvs(ctx context.Context) (*entity.Envs, error) {
 
 	if val, ok := projectCfg.LockedEnvsNames[projectCfg.Environment]; ok && val {
 		fmt.Println(ui.Bold(ui.RedText("Protected Environment Detected!").String()))
-		err := ui.PromptConfirm("Press Enter to Confirm Action")
+		confirm, err := ui.PromptYesNo("Continue fetching variables?")
 		if err != nil {
 			return nil, err
 		}
+		if !confirm {
+			return nil, err
```
JakeCooper, line 37: If not confirm should we return `nil, nil`?
```suggestion
			return nil, nil
```

controller/envs.go
```diff
@@ -64,3 +67,44 @@ func (c *Controller) SaveEnvsToFile(ctx context.Context) error {
...
+		return nil, err
+	}
+	if val, ok := projectCfg.LockedEnvsNames[projectCfg.Environment]; ok && val {
+		fmt.Println(ui.Bold(ui.RedText("Protected Environment Detected!").String()))
+		confirm, err := ui.PromptYesNo("Continue updating variables?")
+		if err != nil {
+			return nil, err
+		}
+		if !confirm {
+			return nil, err
```
JakeCooper, line 83: Same here
```suggestion
			return nil, nil
```

controller/envs.go
```diff
@@ -64,3 +67,44 @@ func (c *Controller) SaveEnvsToFile(ctx context.Context) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	pluginID := ""
+	for _, p := range project.Plugins {
+		if p.Name == "env" {
+			pluginID = p.ID
+		}
+	}
```
JakeCooper, line 97: Should we wrap this into a `getPlugins` function?

gateway/envs.go
```diff
@@ -52,3 +51,30 @@ func (g *Gateway) GetEnvsWithProjectToken(ctx context.Context) (*entity.Envs, er
...
+	gqlReq.Var("envs", req.Envs)
+
+	var resp struct {
+		Envs *entity.Envs `json:"updateEnvsForPlugin"`
+	}
+	if err := g.gqlClient.Run(ctx, gqlReq, &resp); err != nil {
+		return nil, err
+	}
+
+	return &entity.Envs{}, nil
```
JakeCooper, line 79: should this return resp.envs not an empty entity?

JakeCooper (approved):
LGTM with a few small nits

cmd/design.go
gschier: Ah ya, I should change that one to fmt. Yes, `println` is a top-level function that I use all the time for debugging, but I don't think it's recommended practice.

controller/envs.go
gschier: Leaning towards _no_ for now, since it wouldn't save on any code. Also, might need both project and plugins at the same time in some places. Can factor out later if needed.

JakeCooper: SG

gateway/envs.go
gschier: Oh, nice catch! Had it like that because it was failing to deserialize without https://github.com/railwayapp/mono/pull/905

PR #83, opened 2021-04-02
Only get env variables when updating

gschier:
Before, the CLI was getting all plugin variables (pg, redis, etc), then pushing them all to the `env` plugin. This PR fixes this so that it only pulls vars from the `env` plugin.

### DIFF
--- a/controller/envs.go
+++ b/controller/envs.go
@@ -110,7 +110,7 @@ func (c *Controller) UpdateEnvsForEnvPlugin(ctx context.Context, envs *entity.En
 }
 
 func (c *Controller) GetEnvsForEnvPlugin(ctx context.Context) (*entity.Envs, error) {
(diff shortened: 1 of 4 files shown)


controller/envs.go
```diff
@@ -108,3 +108,49 @@ func (c *Controller) UpdateEnvs(ctx context.Context, envs *entity.Envs) (*entity
 		Envs:          envs,
 	})
 }
+
+func (c *Controller) GetEnvsForEnvPlugin(ctx context.Context) (*entity.Envs, error) {
+	// Get envs through production token if it exists
```
coffee-cup, line 113:
```suggestion
	// Get envs through project token if it exists
```

coffee-cup (approved):

JakeCooper (approved):

PR #85, opened 2021-04-07
railway logs command

JakeCooper:
`railway logs` will live tail logs
`railway logs -n 100` will get the last 100 logs (or total logs if less than n)


main.go
```diff
@@ -182,6 +182,12 @@ func init() {
 		RunE:  contextualize(handler.Up, handler.Panic),
 	})
 
+	addRootCmd(&cobra.Command{
+		Use:   "logs",
+		Short: "Get app logs",
```
queer, line 187:
```suggestion
		Short: "View the most-recent deploy's logs",
```

controller/logs.go
```diff
@@ -0,0 +1,62 @@
...
+func (c *Controller) GetActiveDeploymentLogs(ctx context.Context, numLines int32) error {
+	projectID, err := c.cfg.GetProject()
+	if err != nil {
+		return err
+	}
+	environmentID, err := c.cfg.GetEnvironment()
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Fetching latest deployment...")
```
coffee-cup, line 20: Should we print this or use the loading train indicator?

main.go
```diff
@@ -176,11 +176,22 @@ func init() {
...
 	})
 
-	addRootCmd(&cobra.Command{
+	upCmd := &cobra.Command{
 		Use:   "up",
 		Short: "Upload and deploy project from the current directory",
 		RunE:  contextualize(handler.Up, handler.Panic),
-	})
+	}
+	upCmd.Flags().BoolP("detach", "d", false, "--detatch")
```
coffee-cup, line 184: I can't seem to find where this is used. Should it be removed?

controller/logs.go
JakeCooper: I kinda like just mentioning what's actually going on

JakeCooper: Can use the train at the end maybe? It's REALLY quick tho

main.go
JakeCooper: Removed. It's used in the tail up PR

controller/logs.go
coffee-cup: Hmm, if its really quick is it needed? Might be weird for up if you `railway up` and then it shows `Fetching latest deployment`

JakeCooper: Eh that's true

main.go
```diff
@@ -182,6 +182,14 @@ func init() {
 		RunE:  contextualize(handler.Up, handler.Panic),
 	})
 
+	logsCmd := &cobra.Command{
+		Use:   "logs",
+		Short: "View the most-recent deploy's logs",
+		RunE:  contextualize(handler.Logs, handler.Panic),
+	}
+	logsCmd.Flags().Int32P("num_lines", "n", 0, "--lines")
```
gschier, line 190: I think you have the args a bit confused

```suggestion
	logsCmd.Flags().Int32P("lines", "n", 0, "output a specific number of lines")
```

controller/logs.go
```diff
@@ -0,0 +1,60 @@
...
+	environmentID, err := c.cfg.GetEnvironment()
+	if err != nil {
+		return err
+	}
+	deployments, err := c.gtwy.GetDeploymentsForEnvironment(ctx, projectID, environmentID)
+	if err != nil {
+		return err
+	}
+
+	latestDeploy := deployments[0]
```
gschier, line 25: Should we be checking `len(deployments)` here?

Might also be useful to have a `GetLatestDeploymenForEnvironment` fn later on with this logic contained.

controller/logs.go
```diff
@@ -0,0 +1,60 @@
...
+	// Streaming
+	prevIdx := 0
+	for {
+		err := func() error {
```
gschier, line 29: Curious what the purpose of the wrapping function here is?

controller/logs.go
```diff
@@ -0,0 +1,60 @@
...
+			if numLines != 0 {
+				// If num is provided do a walkback by n lines to get latest n logs
+				delimiter = int(math.Max(float64(len(partials)-int(numLines)), float64(prevIdx)))
+			}
+			delta := partials[delimiter:nextIdx]
+			if len(delta) == 0 {
+				return nil
+			}
+			fmt.Println(strings.Join(delta, "\n"))
+			prevIdx = nextIdx
```
gschier, line 49: Might be worth having a high-level comment describing the alg used here. This code is pretty hard to understand.

gateway/deployment.go
```diff
@@ -0,0 +1,66 @@
...
+	err := g.authorize(ctx, gqlReq.Header)
+	if err != nil {
+		return nil, err
+	}
+
+	var resp struct {
+		Deployments []entity.Deployment `json:"allDeploymentsForEnvironment"`
+	}
+	if err := g.gqlClient.Run(ctx, gqlReq, &resp); err != nil {
+		return nil, errors.PluginGetFailed
```
gschier, line 35: Should create a new error for this

gateway/deployment.go
```diff
@@ -0,0 +1,66 @@
...
+	err := g.authorize(ctx, gqlReq.Header)
+	if err != nil {
+		return nil, err
+	}
+
+	var resp struct {
+		Deployment *entity.Deployment `json:"deploymentById"`
+	}
+	if err := g.gqlClient.Run(ctx, gqlReq, &resp); err != nil {
+		return nil, errors.PluginGetFailed
```
gschier, line 63: Same here for the error

gschier (changes requested):
Just a few minor things! Pretty cool that we can whip this up using stuff we already have 👍🏻

controller/logs.go
JakeCooper: Originally, so that I can use return to trigger the deferred sleep, but I pulled it so will remove

controller/logs.go
JakeCooper: Yup will add

gateway/deployment.go
JakeCooper: Oops good catch!

controller/logs.go
JakeCooper: I can slap that into the gateway

coffee-cup: FWIW that query might be useful to have in Backboard. Also might want to check if the latest deploy status is not REMOVED.

JakeCooper: I think it's fine to live in the CLI for now. If we see more usecases for it, we can pull it upwards

Will also filter out removed deployments good call

JakeCooper:
Yea it's neat right? Turns out if you build good data models it's not TOO bad to extend them :D

gschier (approved):
Choo choo!

PR #86, opened 2021-04-07
Tail Up

JakeCooper:
* Adds a fetch + pointer reset for when status changes
* Calls logtail following up


controller/logs.go
```diff
@@ -27,6 +29,7 @@ func (c *Controller) GetActiveDeploymentLogs(ctx context.Context, numLines int32
 	latestDeploy := deployments[0]
 	// Streaming
 	prevIdx := 0
+	status := "SUCCESS"
```
coffee-cup, line 32: Should this come from the `latestDeploy`?

controller/logs.go
```diff
@@ -36,7 +39,14 @@ func (c *Controller) GetActiveDeploymentLogs(ctx context.Context, numLines int32
 			if err != nil {
 				return err
 			}
-			partials := strings.Split(deploy.DeployLogs, "\n")
+			fmt.Println(deploy.Status)
```
coffee-cup, line 42: 🔫

controller/logs.go
JakeCooper: Oh tru

main.go
```diff
@@ -176,11 +176,14 @@ func init() {
...
-	addRootCmd(&cobra.Command{
+	upCmd := &cobra.Command{
 		Use:   "up",
 		Short: "Upload and deploy project from the current directory",
 		RunE:  contextualize(handler.Up, handler.Panic),
-	})
+	}
+	upCmd.Flags().BoolP("detach", "d", false, "--detatch")
+
+	addRootCmd(upCmd)
```
gschier, line 186: FYI `addRootCmd` returns the added cmd so you can add flags, etc. We can probably just get rid of that helper function actually, since it doesn't save use much.

main.go
```diff
@@ -176,11 +176,14 @@ func init() {
...
 	})
 
-	addRootCmd(&cobra.Command{
+	upCmd := &cobra.Command{
 		Use:   "up",
 		Short: "Upload and deploy project from the current directory",
 		RunE:  contextualize(handler.Up, handler.Panic),
-	})
+	}
+	upCmd.Flags().BoolP("detach", "d", false, "--detatch")
```
gschier, line 184:
```suggestion
	upCmd.Flags().BoolP("detach", "d", false, "--detach")
```

gschier (approved):

main.go
```diff
@@ -182,7 +182,13 @@ func init() {
 		Use:   "up",
 		Short: "Upload and deploy project from the current directory",
 		RunE:  contextualize(handler.Up, handler.Panic),
-	})
+	}).Flags().BoolP("detach", "d", false, "--detatch")
```
gschier, line 185: Last arg is the help message.

cmd/up.go
```diff
@@ -18,5 +18,12 @@ func (h *Handler) Up(ctx context.Context, req *entity.CommandRequest) error {
...
 	}
-	return nil
+	detach, err := req.Cmd.Flags().GetBool("detach")
+	if err != nil {
+		return err
+	}
+	if detach {
+		return nil
+	}
+	return h.ctrl.GetActiveDeploymentLogs(ctx, 0)
```
gschier, line 28: Does this mean that `railway up`, by default, streams both build and runtime logs indefinitely after?

cmd/up.go
```diff
@@ -18,5 +18,12 @@ func (h *Handler) Up(ctx context.Context, req *entity.CommandRequest) error {
...
 		ui.StopSpinner(fmt.Sprintf("☁️ Deploy available at %s\n", ui.GrayText(url)))
 	}
-	return nil
+	detach, err := req.Cmd.Flags().GetBool("detach")
+	if err != nil {
+		return err
+	}
+	if detach {
+		return nil
+	}
```
gschier, line 27: Need to stop the spinner at both these points

cmd/up.go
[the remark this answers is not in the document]
gschier: 😆 
image

gschier:
Hmm, I seem to be getting error messages if logs are fetched before the build starts (my guess)? 

image

cmd/up.go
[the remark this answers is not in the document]
JakeCooper: LOL

gschier:
Hmm, also getting some logs printed on the same line. It seems like this happens when there are no new logs to print? It just keeps printing the last one.

image

JakeCooper:
I’ve added a lil loader timeout locally to fix the above

JakeCooper:
RE: Last line yea it’s a small big part of the original log fetcher

Once I merge the refactor I’ll retest but worked on logstreaming

gschier:
> RE: Last line yea it’s a **small big part** of the original log fetcher

😆 

Would be amazing to have some tests for this! 🙏🏻 r

cmd/up.go
JakeCooper: Yes it does

controller/logs.go
```diff
@@ -68,18 +68,26 @@ func (c *Controller) LogsForDeployment(ctx context.Context, req *entity.Deployme
 				return err
 			}
```
gschier, line 69: We should also check `status == "FAILED"` and break out of this

cmd/up.go
gschier: I think I'd vote to only see build logs after `up`. 

My flow is usually to make sure the deploy worked, then go back to writing code. Right now it's difficult to see when it switches from build to deploy then you have to ctrl-C out (which may also not be obvious to less-experienced devs). 

One idea might be to detect when the build succeeded and prompt the user to continue streaming the deploy logs or exit.

controller/logs.go
```diff
@@ -68,18 +68,26 @@ func (c *Controller) LogsForDeployment(ctx context.Context, req *entity.Deployme
...
 			// Current Logs fetched from server
-			currLogs := strings.Split(deploy.DeployLogs, "\n")
+			currLogs := strings.Split(fetchCurrentLogs(deploy), "\n")
 			// Diff logs using the line numbers as references
-			logDiff := currLogs[len(prevLogs)-1 : len(currLogs)-1]
+			idx := int(math.Max(float64(len(idxMp[deploy.Status])-1), 0.0))
```
gschier, line 73: :neckbeard:

controller/logs.go
```diff
@@ -46,17 +45,18 @@ func (c *Controller) LogsForDeployment(ctx context.Context, req *entity.Deployme
...
 		offset = math.Max(float64(len(logLines))-float64(req.NumLines)-1, 0.0)
 	}
 	// Output Initial Logs
-	fmt.Print(strings.Join(logLines[int(offset):], "\n"))
+	fmt.Println(strings.Join(logLines[int(offset):], "\n"))
 	if req.NumLines == 0 {
 		// If no log limit is set, we stream logs
-		prevLogs := strings.Split(deploy.DeployLogs, "\n")
+		idxMp := make(map[string][]string)
+		idxMp[deploy.Status] = strings.Split(deploy.DeployLogs, "\n")
```
queer, line 59: Is this actually correct? Shouldn't it be giving build or deploy logs depending on what the status is?

controller/logs.go
```diff
@@ -68,18 +68,26 @@ func (c *Controller) LogsForDeployment(ctx context.Context, req *entity.Deployme
 				return err
 			}
 			// Current Logs fetched from server
-			currLogs := strings.Split(deploy.DeployLogs, "\n")
+			currLogs := strings.Split(fetchCurrentLogs(deploy), "\n")
 			// Diff logs using the line numbers as references
-			logDiff := currLogs[len(prevLogs)-1 : len(currLogs)-1]
+			idx := int(math.Max(float64(len(idxMp[deploy.Status])-1), 0.0))
+			logDiff := currLogs[idx : len(currLogs)-1]
```
gschier, line 74: Is it possible for `len(currLogs) == 0`? If so, this could panic with `currLogs[0:-1]`.

gschier (commented):
I tested locally, and it the formatting stuff seems to be fixed now! Just a few comments on UX and small stuff 🙌🏻

controller/logs.go
queer: You know, I was happier not knowing that was an emoji.

controller/logs.go
```diff
@@ -76,9 +100,64 @@ func (c *Controller) LogsForDeployment(ctx context.Context, req *entity.Deployme
...
+			if len(logDiff) == 0 {
+				continue
+			}
+			// Output logs
+			fmt.Println(strings.Join(logDiff, "\n"))
 			// Set out walk pointer forward using the newest logs
 			prevLogs = currLogs
+			if deploy.Status != entity.STATUS_BUILDING {
+				break
+			}
```
JakeCooper, line 160: These two functions are SO similar and yet I can't seem to figure out a way to abstract the commonalities into something pretty :(

gschier: Are the only differences `query`, `deploy.BuildLogs` and `STATUS_BUILDING` references? If so, I'd make a new fn with those as args, then have `LogsForBuild` and `LogsForDeploy` call into it.

JakeCooper: Mostly, but they also rely on the closure of having prevlogs memory pointer, as well as this for loop which is broken for build logs but not for deployment logs

Will continue a bit but just wanted to mention/get thoughts

JakeCooper:
Refactored such that we're getting logs for Building and Not Building states. Does a little cutesy 2 state FSM and have unified into "GetLogsForState" functions

cmd/up.go
JakeCooper: I figured the "Build Completed" line is solid for this

My issue is that we fire build completed regardless of outcome, and only fetch the actual deployment logs (Which contains the "Failed to Deploy!" stuff. If we do a prompt, users might not see that

Could see it either way tho. Don't really have a strong opinion so can add a prompt pretty easily

gschier (approved):
Nice lil refactor there 👍🏻

PR #92, opened 2021-04-08
GQL Golang Queries

JakeCooper:
This PR adds a mechanism to do fine grained GQL queries

It's a dependency for granular fetching as part of cooper/tail-up﻿

### DIFF
--- a/lib/gql/gql.go
+++ b/lib/gql/gql.go
@@ -3,12 +3,14 @@ package gql
 import (
 	"context"
 	"encoding/json"
+	"errors"
+	"fmt"
 	"strings"
 )
 
 func AsGQL(ctx context.Context, req interface{}) (*string, error) {
 	// Assume object is a flat keystruct
-	mp := make(map[string]bool)
+	mp := make(map[string]interface{})
 	bytes, err := json.Marshal(req)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


lib/gql/gql.go
```diff
@@ -0,0 +1,28 @@
+package gql
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+)
+
+func AsGQL(ctx context.Context, req interface{}) (*string, error) {
+	// Assume object is a flat keystruct
```
coffee-cup, line 10: I don't think this can always be assumed. For a lot of the queries I think there will be nested objects. Unless maybe I'm misunderstanding what this means

JakeCooper: Was gonna come back in a separate PR and extend it but maybe I just chuck the interface recursion here

Gimme 30m

JakeCooper: Done

JakeCooper:
Now supports nested fields

image

lib/gql/gql.go
```diff
@@ -0,0 +1,42 @@
...
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+func AsGQL(ctx context.Context, req interface{}) (*string, error) {
+	// Assume object is a flat keystruct
+	mp := make(map[string]interface{})
```
JakeCooper, line 13:
```suggestion
	mp := make(map[string]interface{})
```

coffee-cup (approved):

PR #97, opened 2021-04-13
Redeploy after variables change

gschier:
This PR adds logic to redeploy the project after setting/deleting variables. It will only redeploy if a deploy exists, and the last deploy wasn't triggered from `railway up`.

NOTE: I originally planned to move the `up` deployment detection to the GQL endpoint but decided against it so the CLI could have more control over the user experience. I think the ideal case down the road is to store the code used to build/deploy and simply re-trigger that with new variables when it changes. That way, we don't have to care about where the previous deploy came from.

## If last deploy was from GH triggers:

Set variables and show loading while deploy is triggering

image

Display success message when complete

image

## If last deploy was from `up`:

image

### DIFF
--- a/cmd/variables.go
+++ b/cmd/variables.go
@@ -70,6 +70,11 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest) 
 	fmt.Print(ui.Heading(fmt.Sprintf("Updated %s for \"%s\"", strings.Join(updatedEnvNames, ", "), environment.Name)))
 	fmt.Print(ui.KeyValues(*updatedEnvs))
 
+	err = h.ctrl.DeployEnvironmentTriggers(ctx)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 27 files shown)


JakeCooper:
Deployment URL comes back with the response right? Would be nice to have a click that goes directly to the deployment item (similar to the up response)

gschier:
> Deployment URL comes back with the response right? Would be nice to have a click that goes directly to the deployment item (similar to the up response)

Nah, just a boolean, but we could definitely update it to return that.

JakeCooper:
Should be updated regardless. Will do as part of [RAIL-735](https://linear.app/railway/issue/RAIL-735/return-deployed-url)

cmd/variables.go
```diff
@@ -70,6 +70,42 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest)
...
 
+	err = h.maybeRedeployAfterVariablesChange(ctx, environment)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *Handler) maybeRedeployAfterVariablesChange(ctx context.Context, environment *entity.Environment) error {
```
JakeCooper, line 81: lol, can we rename this to `validateAndUpdateDeployment` or something like that. `maybe` just doesn't sit right here

JakeCooper (approved):
Looks good beyond the whole `maybe` thing

cmd/variables.go
gschier: Haha, must be my time with Elm that made me appreciate `maybe`

PR #99, opened 2021-04-17
ProjectConfigNotFound error language updated.

LiamCoop:
railway init was recently separated into railway init & railway link, ProjectConfigNotFound error message updated to reflect this.


JakeCooper (approved):

JakeCooper:
Nice catch and thanks!

PR #101, opened 2021-04-25
One-Off Environments from CLI

JakeCooper:
Added two commands
* `railway environment <newEnvironmentName>`
* `railway run --ephemeral <cmd>`

### DIFF
--- a/cmd/environment.go
+++ b/cmd/environment.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/railwayapp/cli/entity"
 	"github.com/railwayapp/cli/ui"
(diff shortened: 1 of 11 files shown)


cmd/environment.go
```diff
@@ -18,9 +19,23 @@ func (h *Handler) Environment(ctx context.Context, req *entity.CommandRequest) e
 		return err
 	}
 
-	environment, err := ui.PromptEnvironments(project.Environments)
-	if err != nil {
-		return err
+	var environment *entity.Environment
+	if len(req.Args) > 0 {
+		// Create new environment
+		environment, err = h.ctrl.CreateEnvironment(ctx, &entity.CreateEnvironmentRequest{
```
coffee-cup, line 25: This doesn't seem to create an ephemeral environment (like the PR indicates) and rather just creates a normal environment.

JakeCooper: Ya sorry the titles a misnomer

controller/panic.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 func (c *Controller) SendPanic(ctx context.Context, panicErr string, stacktrace string, command string) (bool, error) {
+	fmt.Println(panicErr, stacktrace)
```
JakeCooper, line 13:
```suggestion
```

cmd/run.go
```diff
@@ -61,6 +90,17 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
 
 	printLooksGood()
 
+	if isEphemeral {
```
coffee-cup, line 93: Should we worry about ephemeral environments never being deleted? If the command errors or sigkilled will the ephemeral environment be removed?

cmd/run.go
```diff
@@ -18,6 +18,35 @@ import (
...
+	// Get Current Environment for name
+	environment, err := h.ctrl.GetEnvironment(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Add something to the ephemeral env name
+	if isEphemeral {
+		// Create new environment for this run
+		environment, err = h.ctrl.CreateEnvironment(ctx, &entity.CreateEnvironmentRequest{
```
coffee-cup, line 42: We might want to make these an actual ephemeral environment since it will show differently in the UI and allows you to specify a base environment to "fork" env vars from

JakeCooper: Will do ephemerals + specify current environment

cmd/run.go
JakeCooper: Probs not, but how should we listen to this one?

coffee-cup (approved):

PR #103, opened 2021-04-29
Run command exit code

coffee-cup:
Currently, if the command that `run` is running fails with an exit code, the actual railway run command returns with exit code 0. This isn't great and breaks things like CI.

Ideally we just capture the exit code of the sub command and exit with the same error. But it seems that this is not always the same cross platform.

See

- https://stackoverflow.com/questions/10385551/get-exit-code-go
- https://github.com/golang/go/issues/26539

In go 1.12 they released [ExitCode](https://golang.org/pkg/os/#ProcessState.ExitCode) on ProcessState, but I can not seem to get it to work locally on my Mac. Seems that is the case for others as well.

So as a comprimise, I try to use the exit code if it exists, otherwise I exit with exit code 1.

image

image

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -55,8 +55,13 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
 	catchSignals(cmd)
 
 	err = cmd.Run()
+
 	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


JakeCooper (approved):
Nice pretty simple and should cover us

PR #106, opened 2021-05-04
Show an error if a command is not found or fails in `railway run`

kotx:


cmd/run.go
```diff
@@ -41,6 +41,11 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
 		return errors.CommandNotSpecified
 	}
 
+	if _, err := exec.LookPath(req.Args[0]); err != nil {
+		fmt.Printf("%s is not in $PATH\n", req.Args[0])
+		os.Exit(1)
```
kotx, line 46: This could probably be exit code 127, but that would make programs calling into `railway run` think `railway` is not found AFAIK.

JakeCooper: I think exit(1) is good here. We can change it later if need be. Thanks!

JakeCooper (approved):

PR #112, opened 2021-05-07
Show Dockerbuild + Handle Signal for teardown

JakeCooper:
Previously we didn't show wtf we were doing for the docker build under the hood

Now, we show that such that the user doesn't sit there on this screen. Additionally, we can teardown the container if you hit CTRL + C on the log step now.

image


coffee-cup (approved):
👍  This is probably worth adding to the changelog too

gschier (approved):

Issue #115, opened 2021-05-13
Windows Support


JakeCooper:
Currently Railway lacks windows support, I think mostly due to file system permissions/writing

I'd like to fix this, but currently don't have the cycles to do it. I can get around to it later, but in the meantime, I'll put up $500 in ETH for anybody who resolves this issue in a PR.

JakeCooper:
Resolved by #116 

Confirmation

https://etherscan.io/tx/0x6a172cb78b81a5a0617af4b495d0926a83568ed8bb4b3a823560ae3131aec318

PR #116, opened 2021-05-14
Fix CLI on Windows

rishabluthra:
I think I was able to fix the CLI not working on Windows. As windows doesn't really have a defined HOME directory, chose to use APPDATA/Roaming instead (lots of apps like Heroku CLI use this), ran a few tests with a example project and was able to get it working, short video walkthrough here:

https://www.loom.com/share/4ce6188c7406406a837a7710d70284d8

Let me know if there needs to be any changes or want a longer demo.
Thanks

UPDATE: updated to use `os.UserHomeDir`

### DIFF
--- a/configs/main.go
+++ b/configs/main.go
@@ -6,7 +6,6 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
-	"runtime"
 
 	"github.com/railwayapp/cli/constants"
 	"github.com/spf13/viper"
@@ -84,14 +83,16 @@ func New() *Configs {
 	if IsDevMode() {
 		rootConfigPartialPath = ".railway/dev-config.json"
 	}
-	rootConfigPath := path.Join(os.Getenv("HOME"), rootConfigPartialPath)
-	
-	if (runtime.GOOS == "windows") {
-		rootConfigPath = path.Join(os.Getenv("APPDATA"), rootConfigPartialPath)
-	}
 
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		panic(err)
+	}
+	
+	rootConfigPath := path.Join(homeDir, rootConfigPartialPath)
+	
 	rootViper.SetConfigFile(rootConfigPath)
-	err := rootViper.ReadInConfig()
+	err = rootViper.ReadInConfig()
 	if os.IsNotExist(err) {
 		// That's okay, configs are created as needed
 	} else if err != nil {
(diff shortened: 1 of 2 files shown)


JakeCooper:
Tagging Issue #115

Will test and check, but can you DM me on Discord your preferred method of payment or put your crypto pubkey in the PR?

configs/main.go
```diff
@@ -84,6 +85,11 @@ func New() *Configs {
 		rootConfigPartialPath = ".railway/dev-config.json"
 	}
 	rootConfigPath := path.Join(os.Getenv("HOME"), rootConfigPartialPath)
+	
+	if (runtime.GOOS == "windows") {
+		rootConfigPath = path.Join(os.Getenv("APPDATA"), rootConfigPartialPath)
+	}
```
gschier, line 91: I think we may be able to use Go's built-in `os.UserHomeDir` to obtain the correct directory https://golang.org/pkg/os/#UserHomeDir

rishabluthra: updated the code to use `os.UserHomeDir`, functionality seemed the same, the config will just be in the `C:\Users\USERNAME\` directory now (vs code also uses this). Also did some very limited testing on Mac (just logging in, displaying a projects' env variables and stuff) and it seemed to work (the .railway folder was in the HOME directory like it's supposed to be)

JordanVlieg: `os.UserHomeDir` returns `%USERPROFILE%` on Windows.  This is generally not the preferred place to install app data, that's what appdata is for.

It is personally a pet peeve of mine when applications put any files in my `%USERPROFILE%` instead of Appdata roaming.

JordanVlieg: I personally would recommend using `os.UserConfigDir` which returns `%APPDATA%` on windows. and `$HOME\.config` on Unix.

coffee-cup: That is a good point. We've discussed this and are going to stick with the current implementation (USERPROFILE) for now since it is basically the equivalent of what we do on Unix at the moment. Then in the future, we can migrate both to `os.UserConfigDir`

coffee-cup: Moving to `os.UserConfigDir` right now would require migrating all existing users, which is out of scope of this PR.

coffee-cup (approved):
Thank you!

JakeCooper:
Completed. Thank you!

https://etherscan.io/tx/0x6a172cb78b81a5a0617af4b495d0926a83568ed8bb4b3a823560ae3131aec318

PR #118, opened 2021-05-18
expand open command

FarazPatankar:
This PR closes [RAIL-826](https://linear.app/railway/issue/RAIL-826/expand-open-command-for-cli).

- `railway open` now opens your project dashboard.
- `railway open metrics` will open the metrics page. Aliased as `m`.
- `railway open settings` will open the settings page. Aliased as `s`.
- `railway open live` will open the deployed application. Aliased as `l`.

I've also left some questions in my self-review of the PR.

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -16,5 +16,27 @@ func (h *Handler) Open(ctx context.Context, req *entity.CommandRequest) error {
 		return err
 	}
 
+	if (req.Cmd.Use != "open") {
+		return h.ctrl.OpenProjectPathInBrowser(ctx, projectId, environmentId, req.Cmd.Use)
+	}
+
 	return h.ctrl.OpenProjectInBrowser(ctx, projectId, environmentId)
+}
+
+func (h *Handler) OpenApp(ctx context.Context, req *entity.CommandRequest) error {
+	projectId, err := h.cfg.GetProject()
+	if err != nil {
+		return err
+	}
+	environmentId, err := h.cfg.GetEnvironment()
+	if err != nil {
+		return err
+	}
+
+	deployment, err := h.ctrl.GetLatestDeploymentForEnvironment(ctx, projectId, environmentId)
+	if err != nil {
+		return err
+	}
+
+	return h.ctrl.OpenStaticUrlInBrowser(deployment.StaticUrl)
 }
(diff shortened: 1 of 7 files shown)


cmd/open.go
```diff
@@ -16,5 +16,27 @@ func (h *Handler) Open(ctx context.Context, req *entity.CommandRequest) error {
 		return err
 	}
 
+	if (req.Cmd.Use != "open") {
+		return h.ctrl.OpenProjectPathInBrowser(ctx, projectId, environmentId, req.Cmd.Use)
```
FarazPatankar, line 20: If someone does `railway open potato`, it will simply open their dashboard. Is that okay or do we want to block unknown subcommand?

cmd/open.go
```diff
@@ -16,5 +16,27 @@ func (h *Handler) Open(ctx context.Context, req *entity.CommandRequest) error {
...
+	}
+
 	return h.ctrl.OpenProjectInBrowser(ctx, projectId, environmentId)
 }
+
+func (h *Handler) OpenApp(ctx context.Context, req *entity.CommandRequest) error {
```
FarazPatankar, line 26: Is there a better way to do this? I am fetching the latest deployment and then using the `staticUrl` from within it.

gateway/project.go
```diff
@@ -242,10 +242,18 @@ func (g *Gateway) OpenProjectInBrowser(projectID string, environmentID string) e
...
 func (g *Gateway) OpenProjectDeploymentsInBrowser(projectID string) error {
 	return browser.OpenURL(g.GetProjectDeploymentsURL(projectID))
 }
 
 func (g *Gateway) GetProjectDeploymentsURL(projectID string) string {
 	return fmt.Sprintf("%s/project/%s/deployments?open=true", GetRailwayUrl(), projectID)
 }
+
+func (g *Gateway) OpenStaticUrlInBrowser(staticUrl string) error {
+	return browser.OpenURL("https://" + staticUrl)
```
FarazPatankar, line 258: Should this be a different function that accepts partial URLs like this and then returns the whole thing or are we okay with it being done inline given it's only one case?

cmd/open.go
coffee-cup: I think it would make sense to show help for open command in this case

cmd/open.go
coffee-cup: I think that is the best way

gateway/project.go
coffee-cup: I think it is okay if it is inline. Might be better to use `Sprintf` instead of appending though.

JakeCooper (approved):

PR #121, opened 2021-05-21
Unify Auth Token Authorization

JakeCooper:
The following PR unifies the way we do authentication/authorization in the CLI

g.authorize will now provide the token if it's used for any command﻿

### DIFF
--- a/gateway/main.go
+++ b/gateway/main.go
@@ -25,7 +25,7 @@ func (g *Gateway) authorize(ctx context.Context, header http.Header) error {
 	}
 	header.Add("authorization", fmt.Sprintf("Bearer %s", user.Token))
 	header.Add("x-source", CLI_SOURCE_HEADER)
-	if g.cfg.RailwayProductionToken == "" {
+	if g.cfg.RailwayProductionToken != "" {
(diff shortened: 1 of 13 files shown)


gateway/main.go
```diff
@@ -26,15 +25,9 @@ func (g *Gateway) authorize(ctx context.Context, header http.Header) error {
...
 	header.Add("authorization", fmt.Sprintf("Bearer %s", user.Token))
 	header.Add("x-source", CLI_SOURCE_HEADER)
-	return nil
-}
-
-func (g *Gateway) setProjectToken(ctx context.Context, req *gql.Request) error {
 	if g.cfg.RailwayProductionToken == "" {
-		return errors.ProductionTokenNotSet
+		header.Add("project-access-token", g.cfg.RailwayProductionToken)
 	}
```
coffee-cup, line 30: This adds the token to the header if it empty. I think it should be the other way around

JakeCooper: Wow, odd that it worked then... I suppose it just used my creds will test with a deployment I'm not linked' to

coffee-cup (approved):

PR #129, opened 2021-06-07
Docker tags cannot contain upper-case letters, so lower-case them

gschier:
Upper-case project names were generating upper-case Docker tags, which aren't allowed by docker.

image

After this change:

image


JakeCooper (approved):
Huh

PR #130, opened 2021-06-08
Custom GraphQL Client

gschier:
This PR pulls in a simplified version of the GraphQL client library we were using, so we can extend it to handle errors and things more easily in the future (eg. handling 401 errors).


gateway/main.go
```diff
@@ -14,48 +18,136 @@ const (
...
+		return fmt.Errorf("Response not successful status=%d", res.StatusCode)
 	}
+
+	gr := &GQLResponse{
+		Data: resp,
+	}
+	if err := json.NewDecoder(&buf).Decode(&gr); err != nil {
+		return errors.Wrap(err, "decoding response")
+	}
+	if len(gr.Errors) > 0 {
```
JakeCooper, line 143: Do we want to only return one? Wondering if there's a case here to return all of em joined

JakeCooper (approved):
Looks great. Smolnits

gateway/main.go
gschier: Could? I just pulled this from the old client we were using. Any suggestions welcome

gateway/main.go
```diff
@@ -14,48 +18,137 @@ const (
...
+	}
+
+	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s/graphql", GetHost()), &requestBody)
+	if err != nil {
+		return err
+	}
+
+	req = req.WithContext(ctx)
+	req.Header = r.header
+	req.Header.Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
gschier: Ya, content-type is the request format, which will always be JSON, since even an empty body will be `{variables, query}`

gateway/main.go
JakeCooper: Oh very fair

Ya I'd probably err on the side of doing a join and returning one bigass error, just so we don't bury context

gschier: Hmm, we never actually show these errors anywhere right now but I updated it for later

PR #131, opened 2021-06-08
Exit 1 on error and various tweaks

gschier:
- If any command returns `err`, then `exit(1)`
- To ensure we don't `init` as guest, first try to fetch the user (feels hacky, suggestions welcome)
- Highlight pairing code in magenta for login flow
- Don't `recover()` if running the dev version of the CLI for easier debugging


main.go
```diff
@@ -33,7 +33,13 @@ func addRootCmd(cmd *cobra.Command) *cobra.Command {
 func contextualize(fn entity.HandlerFunction, panicFn entity.PanicFunction) entity.CobraFunction {
 	return func(cmd *cobra.Command, args []string) error {
 		ctx := context.Background()
+
 		defer func() {
+			// Let CLI fail if we're runnign the dev version
+			if !constants.IsDevVersion() {
+				return
+			}
```
JakeCooper, line 41: Hmmm, should we be doing this? At the bare minimum would be great to have a comment why

cmd/init.go
```diff
@@ -181,9 +181,11 @@ func (h *Handler) Init(ctx context.Context, req *entity.CommandRequest) error {
 		return h.Link(ctx, req)
 	}
 
-	isLoggedIn, _ := h.ctrl.IsLoggedIn(ctx)
-
-	if !isLoggedIn {
+	// Since init can be called by guests, ensure we can fetch a user first before calling. This prevents
+	// us accidentally creating a temporary (guest) project if we have a token locally but our remote
+	// session was deleted.
+	_, err := h.ctrl.GetUser(ctx)
```
JakeCooper, line 187: From reading this code, it looks like we're blocking init for all users no?

Just sanity checking logic here

constants/version.go
```diff
@@ -3,3 +3,7 @@ package constants
 const VersionDefault = "Piped into LDflags on build. You are probably running Railway CLI from source."
 
 var Version string = VersionDefault
+
+func IsDevVersion() bool {
```
JakeCooper, line 7: Swear to get we have this or an equivalent somewhere, but I might have removed it....

main.go
gschier: Will add comment to this. Basically, this will let Go print the panic trace instead of us intercepting and showing the "report to railway" bug flow

gschier: But, the logic is actually backwards, so I'll fix that 😆

gschier: Also, here's a screenshot, for completeness
image

cmd/init.go
gschier: If this returns a user correctly, `err == nil` and we continue on our merry way

JakeCooper: Oh right cause this is the gateway got it

JakeCooper (approved):

PR #135, opened 2021-06-10
environment command improvements

FarazPatankar:
### Allow switching to existing environments by name
switch@2x

### Fix cursor in environment selector
cursor@2x

### DIFF
--- a/cmd/environment.go
+++ b/cmd/environment.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/manifoldco/promptui"
 	"github.com/railwayapp/cli/entity"
 	"github.com/railwayapp/cli/ui"
 )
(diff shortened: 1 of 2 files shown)


cmd/environment.go
```diff
@@ -21,17 +22,30 @@ func (h *Handler) Environment(ctx context.Context, req *entity.CommandRequest) e
...
+
+		// Look for existing environment with name
+		for _, projectEnvironment := range project.Environments {
+			if name == projectEnvironment.Name {
+				environment = projectEnvironment
+			}
+		}
+
+		if (environment != nil) {
+			fmt.Printf("%s Environment: %s\n", promptui.IconGood, ui.BlueText(environment.Name))
```
coffee-cup, line 35: nit: The promptui stuff should maybe be added to the ui package

cmd/environment.go
```diff
@@ -21,17 +22,30 @@ func (h *Handler) Environment(ctx context.Context, req *entity.CommandRequest) e
...
+		} else {
+			// Create new environment
+			environment, err = h.ctrl.CreateEnvironment(ctx, &entity.CreateEnvironmentRequest{
+				Name:      name,
+				ProjectID: project.Id,
+			})
+			if err != nil {
+				return err
+			}
+			fmt.Println("Created Environment ✅\nEnvironment: ", ui.BlueText(ui.Bold(name).String()))
```
coffee-cup, line 45: Should this check emoji also be the prompui one?

coffee-cup (approved):

cmd/environment.go
AndreiJirohHaliliDev2006: Probably, but may not work on terminals that don't support emojis, but let's assume it's fine.

cmd/environment.go
FarazPatankar: Will look into this in a separate PR as there seem to be other instances of it being used as well.

PR #139, opened 2021-06-11
Consolidate selection prompts into helper

gschier:
This PR consolidates the prompt selection logic into two helper `selectString` and `selectCustom`.

- All selections now support filtering and show 10 items
- All selections now include a caret for the active item

CleanShot 2021-06-11 at 09 54 51

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -17,13 +17,18 @@ func (h *Handler) Add(ctx context.Context, req *entity.CommandRequest) error {
 	if err != nil {
 		return err
 	}
-	selection, err := ui.PromptPlugins(plugins)
+	selectedPlugin, err := ui.PromptPlugins(plugins)
 	if err != nil {
 		return err
 	}
+	ui.StartSpinner(&ui.SpinnerCfg{
+		Message: fmt.Sprintf("Adding %s plugin", selectedPlugin),
+	})
+	defer ui.StopSpinner("")
+
 	plugin, err := h.ctrl.CreatePlugin(ctx, &entity.CreatePluginRequest{
 		ProjectID: projectCfg.Project,
-		Plugin:    selection,
+		Plugin:    selectedPlugin,
 	})
 	if err != nil {
 		return err
--- a/ui/prompt.go
+++ b/ui/prompt.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 
 	"github.com/manifoldco/promptui"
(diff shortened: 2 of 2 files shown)


cmd/add.go
```diff
@@ -17,13 +17,18 @@ func (h *Handler) Add(ctx context.Context, req *entity.CommandRequest) error {
 	if err != nil {
 		return err
 	}
-	selection, err := ui.PromptPlugins(plugins)
+	selectedPlugin, err := ui.PromptPlugins(plugins)
 	if err != nil {
 		return err
 	}
+	ui.StartSpinner(&ui.SpinnerCfg{
```
gschier, line 24: Bonus tweak!

ui/prompt.go
```diff
@@ -54,24 +50,21 @@ func hasTeams(projects []*entity.Project) bool {
...
-			if project.Team != nil {
-				tm[*project.Team] = true
+			if project.Team == nil {
+				continue
+			}
+
+			// Ensure teams are only appended once by checking teamCheck
+			if _, hasSeenTeam := teamCheck[*project.Team]; !hasSeenTeam {
+				teams = append(teams, *project.Team)
+				teamCheck[*project.Team] = true
```
coffee-cup, line 63: I'm surprised Go doesn't have an Object.values like way to convert map to list of values

ui/prompt.go
```diff
@@ -282,3 +236,59 @@ func validatorRequired(errorMsg string) func(s string) error {
...
+			index:     i,
+		})
+	}
+
+	options := &promptui.Select{
+		Label: fmt.Sprintf("Select %s", label),
+		Items: wrappedItems,
+		Size:  10,
+		Templates: &promptui.SelectTemplates{
+			Active:   fmt.Sprintf(`%s {{ . | underline }}`, promptui.IconSelect),
```
coffee-cup, line 281: Is the custom triangle caret required here?

coffee-cup (approved):
Can't wait for this 😍

ui/prompt.go
JakeCooper: I am not at all surprised by this

https://twitter.com/JustJake/status/1395590075824541697

ui/prompt.go
JakeCooper: I think it's covered by the promptui.IconSelect element

coffee-cup: oh 🤦  I missed that. Carry on!

ui/prompt.go
gschier: Once Generics are in, this should be possible

PR #143, opened 2021-06-14
Send Version with Panic

JakeCooper:
Sends CLI version to Telemetry Endpoint

### DIFF
--- a/controller/user.go
+++ b/controller/user.go
@@ -20,7 +20,7 @@ import (
 
 const (
 	baseRailwayURL   string = "https://railway.app"
-	baseStagingURL   string = "https://railway-staging.app"
+	baseStagingURL   string = "https://railway-web-git-cooper-magic-port-railwayapp.vercel.app"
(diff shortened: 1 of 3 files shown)


controller/user.go
```diff
@@ -20,7 +20,7 @@ import (
 
 const (
 	baseRailwayURL   string = "https://railway.app"
-	baseStagingURL   string = "https://railway-staging.app"
+	baseStagingURL   string = "https://railway-web-git-cooper-magic-port-railwayapp.vercel.app"
```
coffee-cup, line 23: 🔫

coffee-cup (approved):

controller/user.go
JakeCooper: Ah piss nice catch

JakeCooper, line 23:
```suggestion
	baseStagingURL   string = "https://railway-staging.app"
```

PR #147, opened 2021-07-02
fetch all starter templates

FarazPatankar:
As of now, we fetch starter templates from the [`featured.json`](https://github.com/railwayapp/starters/blob/master/featured.json) file in the starters repository. This requires us to manually update this file whenever a new starter is added.

This PR changes it so that the CLI uses the same API that the frontend does.


cmd/init.go
```diff
@@ -44,15 +46,55 @@ func (h *Handler) initNew(ctx context.Context, req *entity.CommandRequest) error
 }
 
 func (h *Handler) initFromTemplate(ctx context.Context, req *entity.CommandRequest) error {
-	// Prompt for a template to use
+	ui.StartSpinner(&ui.SpinnerCfg{
+		Message: "Fetching starter templates",
+	})
```
JakeCooper, line 51: Sanity check: this doesn't return an error right?

entity/starter.go
```diff
@@ -0,0 +1,14 @@
+package entity
+
+type Starter struct {
+	Title 			string 		`json:"title"`
+	Url 				string 		`json:"url"`
+	Source 			string 		`json:"source"`
+}
```
JakeCooper, line 7: Formatting. Should install gopls on your IDE

JakeCooper (approved):
Appears reasonable just a few nits

PR #148, opened 2021-07-09
Support Environment flag for run

JakeCooper:
This PR adds support for environment flags on railway run

`railway run --environment=production <cmd>`﻿

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


gschier (approved):
Looks good! Just realized from looking at an earlier commit here that we don't run golint on this repo, which we should probably add.

gschier:
Oh nvm, we do lint. Was misinterpreting the diff on my phone.

PR #150, opened 2021-07-27
Support --environment for Up + Fix Exit Status'

JakeCooper:
* Supports --environment for up
* Fixed --environment for run to remove the --environment command wherever it is (Allows you to just run it however you'd like)
* Fixed exit statuses to actually show the error message instead of swallowing


cmd/open.go
```diff
@@ -11,17 +11,17 @@ func (h *Handler) Open(ctx context.Context, req *entity.CommandRequest) error {
 	if err != nil {
 		return err
 	}
-	environmentId, err := h.cfg.GetEnvironment()
+	environmentId, err := h.cfg.GetCurrentEnvironment()
```
coffee-cup, line 14: Any reason most of the commands call `GetEnvironment()` as opposed to `GetEnvironment` to use the environment from flag if it exists?

JakeCooper: I don't follow? if you're asking why GetEnvironment doesn't just use the flags cause then we'd have to alter it to take the Cmd as a param and when I did it it was nasty

Still not thritted with this implementation but at least it's clear

coffee-cup (approved):

PR #156, opened 2021-09-23
Feature to up a specific directory

Sunney-X:


JakeCooper:
Awesome! Any shot you could put it as part of the request entity, instead of an additional param?

Other than that looks good and will merge after the above!

Sunney-X:
> Awesome! Any shot you could put it as part of the request entity, instead of an additional param?
> 
> Other than that looks good and will merge after the above!

Hey, could you elaborate on what you mean exactly? I've integrated the param as it is in other commands.

controller/up.go
```diff
@@ -89,9 +89,9 @@ func compress(src string, buf io.Writer) error {
 	return nil
 }
 
-func (c *Controller) Upload(ctx context.Context, req *entity.UploadRequest) (*entity.UpResponse, error) {
+func (c *Controller) Upload(ctx context.Context, req *entity.UploadRequest, src string) (*entity.UpResponse, error) {
```
JakeCooper, line 92: This req is a Upload Request, with a couple properties on it

What I mean is add a property along the lines of "RootDir" and pass it into the function as such, vs creating a new parameter

Sunney-X:
Hey, @JakeCooper your recommendation is done, check out the changes!

JakeCooper (approved):
Awesome thank you!

PR #158, opened 2021-10-04
Ability to Delete Project from CLI

ndneighbor:
Added `railway delete` 

If you provide a project uuid, it will delete the project directly- if you don't it will allow you to browse your account's projects to delete.

Help appreciated on improving the copy on error/success.

Thanks to @TaroballzChen for spearheading this improvement, all I did was debugging.

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -15,9 +15,7 @@ func (h *Handler) Delete(ctx context.Context, req *entity.CommandRequest) error 
 		arg := req.Args[0]
 
 		if uuid.IsValidUUID(arg) {
-			fmt.Printf("Looks good")
 			project, err := h.ctrl.GetProject(ctx, arg)
-			fmt.Printf(project.Id)
 			if err != nil {
 				return err
 			}
@@ -33,8 +31,6 @@ func (h *Handler) Delete(ctx context.Context, req *entity.CommandRequest) error 
 		return h.ctrl.DeleteProject(ctx, project.Id)
 	}
 
-	h.Unlink(ctx, req)
-
 	isLoggedIn, err := h.ctrl.IsLoggedIn(ctx)
 	if err != nil {
 		return err
@@ -49,7 +45,6 @@ func (h *Handler) Delete(ctx context.Context, req *entity.CommandRequest) error 
 }
 
 func (h *Handler) deleteFromAccount(ctx context.Context, req *entity.CommandRequest) error {
-	fmt.Printf("Looks good")
 	projects, err := h.ctrl.GetProjects(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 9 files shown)


cmd/delete.go
```diff
@@ -0,0 +1,80 @@
...
+		return err
+	}
+
+	if isLoggedIn {
+		return h.deleteFromAccount(ctx, req)
+	} else {
+		return h.deleteFromID(ctx, req)
+	}
+
+}
```
JakeCooper, line 45: Delete should use a prompt which requires you to type in the name of the project

ndneighbor: Good point- will be sure to add that to the flow.

TaroballzChen: Thank for your attention

cmd/delete.go
```diff
@@ -0,0 +1,80 @@
...
+	isLoggedIn, err := h.ctrl.IsLoggedIn(ctx)
+	if err != nil {
+		return err
+	}
+
+	if isLoggedIn {
+		return h.deleteFromAccount(ctx, req)
+	} else {
+		return h.deleteFromID(ctx, req)
+	}
```
JakeCooper, line 43: Don't need the else given this is the last function

```suggestion
	return h.deleteFromID(ctx, req)
```

cmd/delete.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/railwayapp/cli/uuid"
+)
+
+func (h *Handler) Delete(ctx context.Context, req *entity.CommandRequest) error {
+	user, err := h.ctrl.GetUser(ctx)
+	if err != nil {
+		return err
+	}
+	if user.Has2FA {
+		fmt.Printf("Your account has 2FA enabled, you must delete your project on the Dashboard.")
```
FarazPatankar, line 18: Maybe add a link to `railway.app/dashboard` which they can just click?

cmd/delete.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (h *Handler) deleteFromID(ctx context.Context, req *entity.CommandRequest) error {
+	projectID, err := ui.PromptText("Enter your project id")
+	if err != nil {
+		return err
+	}
+
+	project, err := h.ctrl.GetProject(ctx, projectID)
+	print("Looks good")
```
FarazPatankar, line 94: Format/remove?

cmd/delete.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (h *Handler) deleteFromAccount(ctx context.Context, req *entity.CommandRequest) error {
+	projects, err := h.ctrl.GetProjects(ctx)
+	if err != nil {
+		return err
+	}
+
+	if len(projects) == 0 {
+		fmt.Printf("No Projects could be deleted.")
```
FarazPatankar, line 62: No projects found?

FarazPatankar (approved):
A few comments about the copy but looks good otherwise.

One case I am not sure if we've covered is, does `getProjects` only return the projects the user is an `admin` of? If not, what happens if they try to delete a project where they're only a member and what error do they see (if any)?

ndneighbor:
@FarazPatankar that might be worth testing, will get to that tomorrow

ndneighbor:
Update, the delete won't go through if you are not authorized to do so- working as intended :)

PR #171, opened 2021-11-02
Attempt to use Label Upstream

JakeCooper:
- Attempts to use the label upstream

Also created a PR upstream to pretty print into the [upstream action](https://github.com/jesusvasquez333/verify-pr-label-action/pull/31)

### DIFF
--- a/.github/workflows/label-check.yml
+++ b/.github/workflows/label-check.yml
@@ -8,7 +8,8 @@ jobs:
   label-check:
     runs-on: ubuntu-latest
     steps:
-      - uses: NathanielHill/check-pr-label-action@v4.0.0
+      - uses: jesusvasquez333/verify-pr-label-action@v1.4.0
         with:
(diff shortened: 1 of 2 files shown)


coffee-cup (approved):
I'm a bit confused what this does

.github/workflows/label-check.yml
```diff
@@ -8,7 +8,8 @@ jobs:
   label-check:
     runs-on: ubuntu-latest
     steps:
-      - uses: NathanielHill/check-pr-label-action@v4.0.0
+      - uses: jesusvasquez333/verify-pr-label-action@v1.4.0
```
JakeCooper, line 11: It alters the check-pr-label action to use the version that's the upstream of the one we currently use (which is an out of date fork)

Previous: https://github.com/NathanielHill/check-pr-label-action

Suggested: https://github.com/jesusvasquez333/verify-pr-label-action

PR #174, opened 2021-11-12
Fix archive/tar: write too long in up command

nebulatgs:
Now copy files into a buffer before streaming them through tar to prevent overwriting the tar stream.


nebulatgs:
messed up the pr

PR #175, opened 2021-11-12
Fix archive/tar: write too long in up command

nebulatgs:


coffee-cup (approved):
Thanks for this! Tested and all seems good 👍

PR #182, opened 2022-01-04
Install binary correctly with NPM

JuanM04:
I replaced `go-npm` with a script that downloads the binaries, untars it, and move them to `./bin` to then be capture by ["bin"](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bin) (degit [uses this method](https://github.com/Rich-Harris/degit/blob/64b80577acf3313b669840f7452800ee8d09fbf3/package.json))

**Important note**: I don't know how to test it with `npm i -g @railway/cli`

### DIFF
--- a/npm-install/config.js
+++ b/npm-install/config.js
@@ -33,7 +33,8 @@ export const CONFIG = {
 export const ARCH_MAPPING = {
   ia32: "386",
   x64: "amd64",
-  arm: "arm",
+  // Re-add when railwayapp/cli#173 gets merged
+  // arm: "arm",
   arm64: "arm64",
 };
 
@@ -44,5 +45,5 @@ export const PLATFORM_MAPPING = {
   darwin: "darwin",
   linux: "linux",
   win32: "windows",
-  freebsd: "freebsd",
+  // freebsd: "freebsd",
 };
--- a/npm-install/config.js
+++ b/npm-install/config.js
@@ -1,0 +1,48 @@
+/**
+ * Global configuration
+ */
+export const CONFIG = {
+  /**
+   * The name of the binary
+   * @type {string}
+   */
+  name: "railway",
+
+  /**
+   * Where to save the unzipped files
+   * @type {string}
+   */
+  path: "./bin",
+
+  /**
+   * The URL to download the binary form
+   *
+   * - `{{arch}}` is one of the Golang achitectures listed below
+   * - `{{bin_name}}` is the name declared above
+   * - `{{platform}}` is one of the Golang platforms listed below
+   * - `{{version}}` is the version number as `0.0.0` (taken from package.json)
+   *
+   * @type {string}
+   */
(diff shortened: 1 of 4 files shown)


JakeCooper:
Juan this is awesome thank you so much!

JakeCooper (approved):

nebulatgs:
seems like you can no longer uninstall the bin though

JakeCooper:
^^That is a good point. What happens when users go to uninstall?

JuanM04:
> What happens when users go to uninstall?

I'm not sure because I don't know how to test it, but since the file is saved inside the package, it should be removed along the contents of the package. When you do `npm uninstall my-package`, the directory `node_modules/my-package` gets deleted, so I think this is the same globally.

[package.json | npm Docs](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bin) says the following:

> Supply a `bin` field in your package.json which is a map of command name to local file name. On install, npm will symlink that file into `prefix/bin` for global installs, or `./node_modules/.bin/` for local installs.
>
> For example, myapp could have this:
> ```json
> { "bin" : { "myapp" : "./cli.js" } }
> ```
>
> So, when you install myapp, it'll create a symlink from the `cli.js` script to `/usr/local/bin/myapp`.

JuanM04:
Hey, are there any updates about this issue?

JakeCooper:
We generally haven't sat down and found a way to test this reliably

Not that it's on you (because we thank you for the PR) but if you'd like to get this in ASAP, is there any way you can author it to like, a dummy NPM package so that we can just validate it, then merge it?

Otherwise I need to do this myself and I'm a tad swamped

JuanM04:
@JakeCooper, oh, yes, I understand. I published for you, it's under `@juanm04/railway-cli`. I installed it with yarn and it seems to work

JakeCooper:
Hmm. Running into this will dig in

CleanShot 2022-02-15 at 14 51 33@2x
.

JakeCooper:
Thank you so much @JuanM04! Can you reach out on Discord? We'd love to give you a badge + some credits!

JuanM04:
> EDIT: Above is node 14. Seems to work on 16+!

Ohhh, maybe it's because it's ESM. Try with the latest version of node 14

> Can you reach out on Discord?

Of course, I'm on the server and I have the same name and profile picture

JakeCooper:
I'm on v.14.19, which seems latest?

CleanShot 2022-02-15 at 15 47 10@2x

JuanM04:
image

It seems that the stream promises were added in v15. You want me to rollback the type = module or should I add a warning for node v14 users?

JakeCooper:
I think if we add an "engine": "node16" to package json that should just error if people try to install it without node16

Or worst case a disclaimer that >16 is required

JuanM04:
Exactly. Something like:

```json
{
  "engines": {
    "node": ">=16.0.0"
  }
}
```

JakeCooper:
Yup that would be perfect! Any shot you can throw that PR up? Otherwise I’ll do it when I’m back in 30-1h

Cheers!

PR #184, opened 2022-01-27
Remove a deprecated function

hrichiksite:
replaced .HasPrefix (https://pkg.go.dev/path/filepath#HasPrefix) with .Match (https://pkg.go.dev/path/filepath#Match)


nebulatgs (changes requested):
Match uses regex, not string matching

JakeCooper:
Covered in another PR. Closing!

PR #187, opened 2022-02-15
Add ignore file

### LINKED ISSUE #181: Don't hardcode paths to ignore in `railway up`
I want to upload my .git folder in order to get the version of my app at runtime using `git rev-parse --short HEAD`. Unfortunately `railway up` does not include `.git`:
https://github.com/railwayapp/cli/blob/a7b020e3a303ae7f839ba9bf7c02c780ac45b88c/controller/up.go#L48
Another use case would probably be modified vendored dependencies in node_modules? Though I believe that's a less common case. 
This issue could be remedied through a `--ignore` flag or a `.railwayignore` instead of automatically ignoring paths, or by just respecting `.gitignore`. 
Anyways, I think cases like these should be supported. I could open a PR for either of the above solutions if needed.

aleksrutins:
Closes #181

This adds support for a .railwayignore file as suggested in #181. If this file is found in the current directory, the CLI will print out `Using ignore file .railwayignore` and use it in addition to `.gitignore` (using the same gitignore parser). If the file isn't found, then it will fall back to using `node_modules/` and `.git/` (the current functionality).

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"time"
 
 	"github.com/railwayapp/cli/entity"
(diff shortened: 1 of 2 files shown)


JakeCooper:
Awesome! Thanks for adding this!

If you wanna reach out on Discord (I'm Cooper) happy to give you "Contributor" status there and also some Railway credits :D

JakeCooper (approved):

PR #193, opened 2022-02-16
Install binary correctly with NPM (again)

JuanM04:
Follow up to #182

This PR adds a static file inside bin that points to the binary to be downloaded by the host computer


JakeCooper (approved):
Let's give err a go!

PR #194, opened 2022-02-16
Verbose mode for `railway up`

### LINKED ISSUE #120: Verbose `railway up`
We should have a `-v` flag for up which relays what's happening internally

aleksrutins:
Closes #120 

This PR adds a `--verbose` flag to the CLI. Currently, only `up` uses it, but it could be used by other commands as well.

The output from #187 would probably be counted as verbose.

Example output:
```
❯ railway up --verbose
💁 Using verbose mode
💁 Loading project configuration
💁 Loading environment
💁 Using environment production
💁 Loading project
💁 Loading services
☁️ Build logs available at https://railway.app/...

==========================
Using Buildpacks!
==========================

...

### DIFF
--- a/ui/text.go
+++ b/ui/text.go
@@ -63,7 +63,7 @@ func AlertInfo(text string) string {
 	return _aurora.Sprintf(GrayText(Bold("💁 %s\n").String()), text)
 }
 
-func AlertVerbose(isVerbose bool, text string) string {
+func VerboseInfo(isVerbose bool, text string) string {
 	if isVerbose {
 		return _aurora.Sprintf(MagentaText("💁 %s\n"), text)
 	} else {
--- a/ui/text.go
+++ b/ui/text.go
@@ -2,10 +2,11 @@ package ui
 
 import (
 	"fmt"
-	_aurora "github.com/logrusorgru/aurora"
 	"math"
 	"sort"
 	"strings"
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by JakeCooper** [comment]
This is awesome! Just going to resolve some issues with npm/yarn based distro on Mac and we can merge this in later today!

File: ui/text.go [hunk 1]
```diff
@@ -62,6 +63,14 @@ func AlertInfo(text string) string {
 	return _aurora.Sprintf(GrayText(Bold("💁 %s\n").String()), text)
 }
 
+func AlertVerbose(isVerbose bool, text string) string {
```
JakeCooper, line 66: Nit: can we call this something like `InfoIfVerbose` or `VerboseInfo`?

The naming to me suggests that something will print, and you can add verbose to get WAY more. Maybe just me. Thoughts?

JakeCooper (approved):
Looks great! Let me know what you wanna do about the nit and I'll merge after you've made the call :D

Thanks!

ui/text.go
aleksrutins: That does make sense. I've renamed it to `VerboseInfo`.

aleksrutins:
On second thought, would `-v` conflict with `--version`?

JakeCooper:
Version I think is it's own command for this reason

PR #202, opened 2022-02-18
fix: npm install for windows

### LINKED ISSUE #201: NPM install is broken for windows
`npm i -g @railway/cli` is broken as it is trying to read the `package.json` version which does not match any of the released versions.

I have made a potential fix here https://github.com/maxs-rose/cli/commit/bd50117f8fa93b80ecb434a00da2b1f8652a3e01 the issue is it will only grab the latest release and that's why I haven't made a pull request for it. 

If we are happy with the idea I can make a pull request for it.

maxs-rose:
Fixes npm install for windows.

closes #201


maxs-rose:
Pulled from the wrong branhc

PR #204, opened 2022-02-18
Support alternate Dockerfile for `run`

### LINKED ISSUE #164: Give `railway run` a specific Dockerfile
Often, you have different Dockerfiles depending on your application's environment, i.e., one for development and another for production. Currently, when running `railway run,` railway picks up only the one named `Dockerfile` and not something like `Dockerfile.dev`. It would be awesome if the user could choose the specific Dockerfile that `railway run` should use :)

aleksrutins:
Closes #164

This PR adds support for an alternate Dockerfile (`Dockerfile.dev`) to `railway run`. If `Dockerfile.dev` is found, it will use that; otherwise, it will fall back to `Dockerfile`.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -146,7 +146,7 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
 	return nil
 }
 
-func (h *Handler) runInDocker(ctx context.Context, pwd string, dockerfile string, envs *entity.Envs) error {
+func (h *Handler) runInDocker(ctx context.Context, pwd, dockerfile string, envs *entity.Envs) error {
 	// Start building the image
 	projectCfg, err := h.ctrl.GetProjectConfigs(ctx)
 	if err != nil {
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -89,13 +89,18 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
 	}
 
 	hasDockerfile := true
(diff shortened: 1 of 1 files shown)


cmd/run.go
```diff
@@ -40,6 +40,12 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
 		return err
 	}
 
+	dockerfileRegex, err := regexp.Compile("--dockerfile=(.*)")
+	dockerfile := "Dockerfile.dev"
```
coffee-cup, line 44: I think this should just default to `Dockerfile`

cmd/run.go
```diff
@@ -90,12 +101,16 @@ func (h *Handler) Run(ctx context.Context, req *entity.CommandRequest) error {
 
 	hasDockerfile := true
 
+	if _, err := os.Stat(fmt.Sprintf("%s/%s", pwd, dockerfile)); os.IsNotExist(err) {
+		dockerfile = "Dockerfile"
+	}
+
 	if _, err := os.Stat(fmt.Sprintf("%s/Dockerfile", pwd)); os.IsNotExist(err) {
```
coffee-cup, line 108: I am seeing errors when running in project that only has an alternativley name Dockerfile. I think it is because it is checking here if `./Dockerfile` exists. We probably want to check `./${dockerfile}` instead.

image

coffee-cup (changes requested):

PR #205, opened 2022-02-18
Add git metadata to deployments with `railway up`

### LINKED ISSUE #178: show commit message and branch when deploying from github actions
It would be really nice to also see the branch name and commit message in the railway dashboard when deploying with github actions. 

I'm using the railway cli as described in https://blog.railway.app/p/github-actions, but right now it only seems shows the branch name and commit message for automated deployments with github hooks:

dashboard

I'm not very familiar with go, but maybe something like this could work:

```go
var branchName = ""
var commitMessage = ""
if os.Getenv("GITHUB_ACTIONS") != "" {
  branchName = os.Getenv("GITHUB_REF_NAME")
  buffer, err := exec.Command("/bin/sh", "git log -1 --pretty=%B").Output()
  if err != nil {
    fmt.Println("error:", err)
  }
  commitMessage = string(buffer)
}

// TODO use the variables and send them to railway dashboard somehow
fmt.Println("branchName:", branchName)
fmt.Println("commitMessage:", commitMessage)
```

What do you think?

aleksrutins:
Closes #178 (at least, it should)

This PR will add git metadata to deployments with `railway up` when deploying a directory with a git repository.

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -67,20 +67,17 @@ func (h *Handler) Up(ctx context.Context, req *entity.CommandRequest) error {
 	fmt.Print(ui.VerboseInfo(isVerbose, "Getting Git information"))
 	gitInfo, err := git.GetAllMetadata(src)
 
-	if !gitInfo.IsRepo {
+	if !gitInfo.IsRepo || err != nil {
 		fmt.Print(ui.VerboseInfo(isVerbose, "No Git repository found"))
-		fmt.Print(err)
 	}
-	{
-		fmt.Print(ui.VerboseInfo(isVerbose, fmt.Sprintf(`Git information:
-   Repository name: %s
-   Branch: %s
-   Latest commit:
-	   Hash: %s
-	   Message: %s
-	   Author: %s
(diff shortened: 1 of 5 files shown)


maxs-rose:
@aleksrutins at least for a very basic repo it gets the correct information on windows :)

image

image

aleksrutins:
Cool! If the repo has an `origin` remote on GitHub, though, `Repository name` is supposed to print out as `user/repo` instead of just the directory name, so I'm not sure if that quite works correctly. The logic for that is the most likely to be non-Windows-compatible.

aleksrutins:
What would the API for this be? That's the only thing blocking me at this point.

coffee-cup:
Hey @aleksrutins,

Apologies for the late reply.

After thinking about this for a while I am not sure we want this in the CLI. The problem I see is if there are unstaged changes and we incorrectly attribute the deploy with the latest commit. This could show a failing build because of some local changes but when deployed from git from the actual commit the deploy would succeed. I guess we could only add the git information if there are no local changes? Thoughts?

aleksrutins:
Yeah, good point - I didn't think of that. I should make it only run if there are no local changes if this is going to get into the main code. I was kind of waiting until the public API docs were released at this point to continue working on this, since that's really the only thing (other than the local changes bit) left.

PR #208, opened 2022-02-25
Service support for up, run, and vars commands

coffee-cup:
A few changes to make working with multi-service projects easier

- `--service {serviceName}` flag support for up, run, and vars commands
- Prompt for service if necessary for run, vars, and var subcommands

image


JakeCooper (approved):
Looks good!

PR #210, opened 2022-03-24
controller/up: properly handle nested ignore files

morgangallant:
Previously, we'd only consider top level ignore files (.gitignore,
.railwayignore), which led to some problems when doing `railway up`
in a monorepo environment.

For example,
```
mg@mbp operand % railway up
🚄 Laying tracks in the clouds... open web/node_modules/core-js-pure/actual/typed-array/reverse.js: too many open files
```

To fix this, we do a pre-scan to find all the relevant ignore
files in the project first, then apply them as necessary.

Some additional cleanup / fixes as well (e.g., closing files, etc).

### DIFF
--- a/controller/up.go
+++ b/controller/up.go
@@ -32,7 +32,7 @@ type ignoreFile struct {
 }
 
 func scanIgnoreFiles(src string) ([]ignoreFile, error) {
-	ret := []ignoreFile{}
+	ignoreFiles := []ignoreFile{}
 
 	if err := filepath.WalkDir(src, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
@@ -62,7 +62,7 @@ func scanIgnoreFiles(src string) ([]ignoreFile, error) {
 				prefix = "" // Handle root dir properly.
 			}
 
-			ret = append(ret, ignoreFile{
+			ignoreFiles = append(ignoreFiles, ignoreFile{
 				prefix: prefix,
 				ignore: igf,
 			})
@@ -73,7 +73,7 @@ func scanIgnoreFiles(src string) ([]ignoreFile, error) {
 		return nil, err
 	}
 
-	return ret, nil
+	return ignoreFiles, nil
 }
 
 func compress(src string, buf io.Writer) error {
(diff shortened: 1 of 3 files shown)


controller/up.go
```diff
@@ -7,38 +7,100 @@ import (
...
+	"node_modules",
+}
+
+type ignoreFile struct {
+	prefix string
+	ignore *gitignore.GitIgnore
+}
+
+func scanIgnoreFiles(src string) ([]ignoreFile, error) {
+	ret := []ignoreFile{}
```
JakeCooper, line 35: nit: ret should be more descriptive

Suggestion: `ignoredFilenames`

controller/up.go
```diff
@@ -7,38 +7,100 @@ import (
...
+}
+
 func compress(src string, buf io.Writer) error {
 	// tar > gzip > buf
 	zr := gzip.NewWriter(buf)
 	tw := tar.NewWriter(zr)
-	ignore, err := gitignore.CompileIgnoreFile(".gitignore")
 
+	// find all ignore files, including those in subdirs
+	ignoreFiles, err := scanIgnoreFiles(src)
```
JakeCooper, line 85: ignoredFilenames, same as the return type

controller/up.go
```diff
@@ -68,6 +135,9 @@ func compress(src string, buf io.Writer) error {
 			return err
 		}
 
+		// close the file to avoid hitting fd limit
+		_ = f.Close()
```
JakeCooper, line 139: if there's an error we should probably alert the user but complete anyways

Something like this (don't know what the API is for file off the top of my head)

```
		err = f.Close()
		if err != nil {
		    fmt.Printf("Failed to add %s", f.name) 
		}
```

JakeCooper (approved):
Looks good just a couple comments and then I'll merge it!

controller/up.go
morgangallant: Changed to `ignoreFiles`, wouldn't want to get an `ignoreFile` object confused w/ some list of file names.

controller/up.go
morgangallant: See above.

controller/up.go
morgangallant: agreed

morgangallant:
99% sure this shouldn't break anything, but does change behaviour of `node_modules` and `.git` slightly.

Prev:
```
rwIgnore, err := gitignore.CompileIgnoreFile(".railwayignore")

	if err != nil {
		rwIgnore, err = gitignore.CompileIgnoreLines(".git/", "node_modules/")
		if err != nil {
			return err
		}
	}
```

e.g., would only ignore these files (at the top level) if `.railwayignore` file not present.

PR #213, opened 2022-04-12
Improved to check project id in "whoami" command

hmmhmmhm:
스크린샷 2022-04-12 오후 1 14 54
I added the code so that we can see the project id in cli.


coffee-cup:
Mentioned this in Discord, but I don't think the `whoami` command should show any project information. Instead I would prefer a `--ids` flag on `railway status` that output the IDs.

Going to close for now, but feel free to make a new PR with update to status command instead.

PR #216, opened 2022-04-22
`railway shell` command

coffee-cup:
`railway shell` command that creates a subshell (using `$SHELL`) with the variables from environment/service loaded and available.

Thanks for the [suggestion wyzlle](https://canary.discord.com/channels/713503345364697088/943650028696768532/967033569203617833)

image

### DIFF
--- a/cmd/shell.go
+++ b/cmd/shell.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime"
 
 	"github.com/railwayapp/cli/entity"
 	"github.com/railwayapp/cli/ui"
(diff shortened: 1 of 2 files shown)


wyzlle:
Are you sure this will work on `windows`?

coffee-cup:
I think so. It will fallback to the `bash` command which should be available in newer versions of windows. It should also work in WSL.

Unfortunately I don't have a Windows machine available to test on.

GrimZZZ-404:
Would be better if it defaults to `cmd` on windows.

coffee-cup:
Changed to fallback to `cmd` which I think makes sense

GrimZZZ-404 (approved):

Milo123459:
What about powershell?

wyzlle:
> What about powershell?

You can always do `railway run powershell`. That would be better than adding a flag in `raliway shell` command imo.

coffee-cup:
Unfortunately, I don't have a Windows machine handly so don't want to complicate the Window logic any more than it is without being able to properly test it. Would be happy to accept a PR after thought to add Powershell support (rather detect what the current shell the user is in, if not using the `$SHELL` variable).

JakeCooper (approved):

gschier (commented):
Truly beautiful ❤️

PR #241, opened 2022-07-10
Upgrade command

cchalop1:
Hi,

I made a first implementation of the upgrade command.

It's not completely finished, the different flags are missing in the build part.

For example like this for brew:

```
go build -o bin/railway "-ldflags "-X main.OriginalInstallationMethod=brew"
```

And finally I'm waiting for your feedback especially on the gender difference between yarn and npm.

thank you
have a good day

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -41,8 +41,8 @@ func getUpdateCommand(installationMethod string) (*exec.Cmd, error) {
 }
 
 func (h *Handler) Upgrade(ctx context.Context, req *entity.CommandRequest) error {
-	currentVersion, _ := h.ctrl.GetLatestVersion()
-	if currentVersion == constants.Version {
+	latestVersion, _ := h.ctrl.GetLatestVersion()
+	if latestVersion == constants.Version {
 		fmt.Printf("\nYou are currently up to date")
 		return nil
 	}
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -55,7 +55,7 @@ func (h *Handler) Upgrade(ctx context.Context, req *entity.CommandRequest) error
 	}
 
 	if h.ctrl.RunUpdateCommand(updateCommand) == nil {
-		fmt.Printf("Error when we try to run upload command")
+		fmt.Printf("Error when we try to run upgrade command")
 		return nil
 	}
 	fmt.Printf("upload run sucressfuly")
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -13,8 +12,6 @@ import (
 	"github.com/railwayapp/cli/entity"
 )
 
-var OriginalInstallationMethod string
-
 func getExectuablePath() string {
 	ex, err := os.Executable()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -126,6 +129,12 @@ func init() {
 		Deprecated: "Please use 'railway variables' instead", /**/
 	})
 
+	addRootCmd(&cobra.Command{
+		Use:   "upgrade",
+		RunE:  contextualize(handler.Upgrade, handler.Panic),
+		Short: "upgrade railway cil version if new one",
```
JakeCooper, line 135:
```suggestion
		Short: "Upgrades the railway cli to the latest version",
```

cmd/upgrade.go
```diff
@@ -0,0 +1,64 @@
...
+	if e != nil {
+		fmt.Printf(e.Error())
+		return nil
+	}
+
+	if h.ctrl.RunUpdateCommand(updateCommand) == nil {
+		fmt.Printf("Error when we try to run upload command")
+		return nil
+	}
+	fmt.Printf("upload run sucressfuly")
```
JakeCooper, line 61:
```suggestion
	fmt.Printf("Upgrade run sucressfuly")
```

cmd/upgrade.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	updateCommand, e := getUpdateCommand(OriginalInstallationMethod)
+	fmt.Println(updateCommand)
+	if e != nil {
+		fmt.Printf(e.Error())
+		return nil
+	}
+
+	if h.ctrl.RunUpdateCommand(updateCommand) == nil {
```
JakeCooper, line 57: != nil?

cmd/upgrade.go
```diff
@@ -0,0 +1,64 @@
...
+		fmt.Printf("Error when we try to run upload command")
```
JakeCooper, line 58:
```suggestion
		fmt.Printf("Error when we try to run upgrade command")
```

cmd/upgrade.go
```diff
@@ -0,0 +1,64 @@
...
+		} else if strings.Contains(currentPath, "yarn") {
+			return exec.Command("yarn", "global", "add", "@railway/cli"), nil
+		}
+	}
+	return nil, errors.New("installation methodes is not recognized")
+}
+
+func (h *Handler) Upgrade(ctx context.Context, req *entity.CommandRequest) error {
+	currentVersion, _ := h.ctrl.GetLatestVersion()
+	if currentVersion == constants.Version {
```
JakeCooper, line 45:
```suggestion
	latestVersion, _ := h.ctrl.GetLatestVersion()
	if latestVersion == constants.Version {
```

cmd/upgrade.go
```diff
@@ -0,0 +1,64 @@
...
+		return exec.Command("curl", "-fsSL", "https://railway.app/install.sh", "|", "sh"), nil
+	} else if installationMethod == "npm" {
+		currentPath := getExectuablePath()
+		if strings.Contains(currentPath, "npm") {
+			return exec.Command("npm", "i", "-g", "@railway/cli"), nil
+		} else if strings.Contains(currentPath, "yarn") {
+			return exec.Command("yarn", "global", "add", "@railway/cli"), nil
+		}
+	}
+	return nil, errors.New("installation methodes is not recognized")
```
JakeCooper, line 40:
```suggestion
	return nil, errors.New("installation method not recognized")
```

JakeCooper (approved):
Looks great! Including the npm/yarn stuff.

Few nits on copy/variables but if you've tested happy to merge it! Thanks for the contribution!

cchalop1:
Hi, thanks for your return

cchalop1:
What remains to be done is to implement the build part

cmd/upgrade.go
```diff
@@ -0,0 +1,64 @@
...
+		return exec.Command("brew", "upgrade", "railway"), nil
+	} else if installationMethod == "curl" {
+		return exec.Command("curl", "-fsSL", "https://railway.app/install.sh", "|", "sh"), nil
+	} else if installationMethod == "npm" {
+		currentPath := getExectuablePath()
+		if strings.Contains(currentPath, "npm") {
+			return exec.Command("npm", "i", "-g", "@railway/cli"), nil
+		} else if strings.Contains(currentPath, "yarn") {
+			return exec.Command("yarn", "global", "add", "@railway/cli"), nil
+		}
```
JakeCooper, line 38: I'm actually not sure how you'd distribute multiple binaries for each of these :/

Frankly, you may want to just go with the `getExecutablePath()` method for homebrew/CLI, and then if you can't detect the installation for upgrade throw

JakeCooper:
Ah shit I see what you mean. I don't know if there's a way to actually ship those multiple binaries without a ton of work.

Added a comment but you may just want to get with the execution path method and then throw if you can't find it for brew/yarn/npm/curl

cchalop1:
Hum ok I will try with only the path of the executable. I thinks is not cover all the case but we can start with this i this it.

cmd/upgrade.go
```diff
@@ -0,0 +1,64 @@
...
+	ex, err := os.Executable()
+	if err != nil {
+		panic(err)
+	}
+	exPath := filepath.Dir(ex)
+	return exPath
+}
+
+func getUpdateCommand(installationMethod string) (*exec.Cmd, error) {
+	if installationMethod == "brew" {
```
JakeCooper, line 28: Sorry, what I meant with the previous comment was to replace this (1/2)

cmd/upgrade.go
```diff
@@ -0,0 +1,64 @@
...
+		return exec.Command("brew", "upgrade", "railway"), nil
+	} else if installationMethod == "curl" {
+		return exec.Command("curl", "-fsSL", "https://railway.app/install.sh", "|", "sh"), nil
+	} else if installationMethod == "npm" {
+		currentPath := getExectuablePath()
+		if strings.Contains(currentPath, "npm") {
```
JakeCooper, line 34: (2/2) entirely with this, such that we can do it based on path detection and not have to change the binaries build input (and thus distribute many different binaries)

JakeCooper:
Couple comments and build looks to be failling. If we can resolve those, would love to get this merged!

PR #245, opened 2022-07-28
Up from project root

coffee-cup:
- Set railway cli version as header on up request
- Up from the project root instead of the current directory

### DIFF
--- a/controller/up.go
+++ b/controller/up.go
@@ -153,7 +153,7 @@ func compress(src string, buf io.Writer) error {
 
 		// must provide real name
 		// (see https://golang.org/src/archive/tar/common.go?#L626)
-		header.Name = filepath.ToSlash(absoluteFile)
+		header.Name = filepath.ToSlash(relativeFile)
 		// size when we first observed the file
 		header.Size = int64(data.Len())
 
--- a/controller/up.go
+++ b/controller/up.go
@@ -88,7 +88,12 @@ func compress(src string, buf io.Writer) error {
 	}
 
 	// walk through every file in the folder
-	err = filepath.WalkDir(src, func(file string, de os.DirEntry, passedErr error) error {
+	err = filepath.WalkDir(src, func(absoluteFile string, de os.DirEntry, passedErr error) error {
+		relativeFile, err := filepath.Rel(src, absoluteFile)
+		if err != nil {
+			return err
+		}
+
 		if passedErr != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


controller/up.go
```diff
@@ -105,16 +110,16 @@ func compress(src string, buf io.Writer) error {
 		}
 
 		for _, igf := range ignoreFiles {
-			if strings.HasPrefix(file, igf.prefix) { // if ignore file applicable
-				trimmed := strings.TrimPrefix(file, igf.prefix)
+			if strings.HasPrefix(relativeFile, igf.prefix) { // if ignore file applicable
+				trimmed := strings.TrimPrefix(relativeFile, igf.prefix)
```
gschier, line 114: Does this still work with relative file paths instead of absolute?

gateway/up.go
```diff
@@ -18,6 +19,13 @@ func constructReq(ctx context.Context, req *entity.UpRequest) (*http.Request, er
 		return nil, err
 	}
 	httpReq.Header.Set("Content-Type", "multipart/form-data")
+
+	version := constants.Version
+	if constants.IsDevVersion() {
+		version = "dev"
```
gschier, line 25: Do we still want to send the version in dev so that we can test things on Backboard that might be using it? Could add an additional header for variant if we need to check dev/staging/prod

gateway/up.go
```diff
@@ -18,6 +19,13 @@ func constructReq(ctx context.Context, req *entity.UpRequest) (*http.Request, er
...
+	}
+	httpReq.Header.Set("X-Railway-Version", version)
```
gschier, line 27: Do we want to send this header on all requests (including GQL)? If so, we can pull the http client out of cli/gateways/main.go into a common client and configure it to add the header on every request.

controller/up.go
coffee-cup: The previous paths were all relative to the current directory so this is just matching that behaviour. The ignore paths are relative (e.g. node_modules) so we need to compare against the path relative to the project root.

gateway/up.go
coffee-cup: As far as I know there is no way to get the version in dev. The version is injected at compile time in CI based off the git tag.

gateway/up.go
coffee-cup: Yes that is a good idea

gateway/up.go
gschier: Oh right, good point. That's too bad but understandable

gschier (approved):
Beautiful <3

PR #249, opened 2022-08-03
Run Homebrew Bump after GoReleaser

JakeCooper:
I don't think you can actually just listen to the publish event, so just sidecar this onto the brew action and run it after GoReleaser publishes it

Also removes a lot of options to just let it infer it automagically

See: https://github.com/dawidd6/action-homebrew-bump-formula#standard-mode

CleanShot 2022-08-03 at 13 46 12@2x


.github/workflows/publish.yml
```diff
@@ -29,6 +29,12 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GH_PAT }}
 
+      - name: Bump Homebrew Core
+      - uses: dawidd6/action-homebrew-bump-formula@v3
+        with:
```
coffee-cup, line 34: I think we also need to add the `version` right?

JakeCooper: I think it'll just pick the ref?

CleanShot 2022-08-03 at 13 51 01@2x

Which is what we do here right?

CleanShot 2022-08-03 at 13 51 20@2x

coffee-cup: Oh perhaps. Not really sure tbh. I was just thinking we do use the version manually here https://github.com/railwayapp/cli/blob/master/.github/workflows/publish.yml#L65.

But can give it a go like this.

PR #265, opened 2022-08-20
Introduce optional replace semantics for `railway variables set` as an alternative to upsert

### LINKED ISSUE #264: Add a flag to the CLI to use the `variablesSetFromObject` mutation instead of `upsertVariablesFromObject`
Hey there, at the moment I am working on writing a custom integration between Doppler and Railway since there is no official solution just yet. So far I have been using the `railway variables set` command (triggered by a webhook) which technically works but has only upsert semantics using the `upsertVariablesFromObject` mutation unlike the `variablesSetFromObject` mutation that the raw editor in the UI uses. As such deletions of variables within Doppler aren't correctly propagated.

I'd like to ask whether you'd be willing to add a flag to the CLI to specifically overwrite all existing variables instead of just upserting. If you are open to this suggestion I might try adding it myself even though I have no experience with Go, but I have a feeling the task is simple enough to learn it that way.

ChristianIvicevic:
In this PR I have added an additional flag `replace` for `variables set` that replaces all variables of a service rather than just upserting them. This is helpful for automation using another single source of truth to push a snapshot of variables; otherwise, old variables have to be deleted one-by-before before setting the current ones.

I tried to reuse as much code as possible and thus renamed functions that were prefixed with `Upsert` in favor of `Update`. Additionally I have intentionally opted out of a short-hand form for the `--replace` flag to ensure nobody triggers it by accident.

Looking forward for feedback on this one.

Closes #264

### DIFF
--- a/cmd/variables.go
+++ b/cmd/variables.go
@@ -63,6 +63,12 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest) 
 		skipRedeploy = false
 	}
 
+	replace, err := req.Cmd.Flags().GetBool("replace")
+	if err != nil {
+		// The flag is optional; default to false.
+		replace = false
+	}
+
 	variables := &entity.Envs{}
 	updatedEnvNames := make([]string, 0)
 
@@ -78,7 +84,7 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest) 
 		updatedEnvNames = append(updatedEnvNames, key)
 	}
 
-	err = h.ctrl.UpsertEnvs(ctx, variables, &serviceName)
+	err = h.ctrl.UpdateEnvs(ctx, variables, &serviceName, replace)
 
 	if err != nil {
 		return err
@@ -89,7 +95,12 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest) 
 		return err
 	}
 
-	fmt.Print(ui.Heading(fmt.Sprintf("Updated %s for \"%s\"", strings.Join(updatedEnvNames, ", "), environment.Name)))
+	operation := "Updated"
+	if replace {
+		operation = "Replaced existing variables with"
+	}
+
+	fmt.Print(ui.Heading(fmt.Sprintf("%s %s for \"%s\"", operation, strings.Join(updatedEnvNames, ", "), environment.Name)))
 	fmt.Print(ui.KeyValues(*variables))
 
 	if !skipRedeploy {
--- a/cmd/variables.go
+++ b/cmd/variables.go
@@ -69,6 +69,23 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest) 
 		replace = false
 	}
 
+	yes, err := req.Cmd.Flags().GetBool("yes")
+	if err != nil {
+		// The flag is optional; default to false.
+		yes = false
+	}
+
+	if replace && !yes {
+		fmt.Println(ui.Bold(ui.RedText(fmt.Sprintf("Warning! You are about to fully replace all your variables for the service '%s'.", serviceName)).String()))
+		confirm, err := ui.PromptYesNo("Continue?")
+		if err != nil {
+			return err
+		}
+		if !confirm {
+			return nil
+		}
+	}
+
 	variables := &entity.Envs{}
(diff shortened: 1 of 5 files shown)


JakeCooper:
Oh wild. Does this work?

ChristianIvicevic:
It does, I tested it with one of my own projects/services. :)
If you want to give it a try yourself, I'd appreciate it just to make sure I didn't make a mistake somewhere.

cmd/variables.go
```diff
@@ -63,6 +63,12 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest)
 		skipRedeploy = false
 	}
 
+	replace, err := req.Cmd.Flags().GetBool("replace")
+	if err != nil {
+		// The flag is optional; default to false.
+		replace = false
+	}
```
JakeCooper, line 70: Thoughts on having a prompt that says "WARNING YOU'RE REPLACING ALL YOUR VARIABLES! ONLY DO THIS IF YOU KNOW WHAT YOU'RE DOING!"

JakeCooper: Like ideally would love a "--replace --yes" flag

gateway/envs.go
```diff
@@ -32,12 +33,18 @@ func (g *Gateway) GetEnvs(ctx context.Context, req *entity.GetEnvsRequest) (*ent
...
 }
 
-func (g *Gateway) UpsertVariablesFromObject(ctx context.Context, req *entity.UpdateEnvsRequest) error {
-	gqlReq, err := g.NewRequestWithAuth(`
+func (g *Gateway) UpdateVariablesFromObject(ctx context.Context, req *entity.UpdateEnvsRequest, replace bool) error {
+	queryName := "upsertVariablesFromObject"
+
+	if replace {
+		queryName = "variablesSetFromObject"
+	}
```
JakeCooper, line 41: Kinda a potential foot gun. I know I gave you the "replace" verbiage but it's one of those things where I'd hate someone to call this function unknowingly, have a release get cut, and then accidently delete all their variables :/

```suggestion
	if replace {
	        // When replacing, use the set query which will blow away all old variables and only set the ones in this query
		queryName = "variablesSetFromObject"
	}
```

ChristianIvicevic: I don't really understand what you are implying with your suggested change. Just a comment for somebody working on the CLI code or something beyond that?

main.go
```diff
@@ -154,6 +154,8 @@ func init() {
 	variablesCmd.AddCommand(variablesSetCmd)
 	variablesSetCmd.Flags().StringP("service", "s", "", "Fetch variables accessible to a specific service")
 	variablesSetCmd.Flags().Bool("skip-redeploy", false, "Skip redeploying the specified service after changing the variables")
+	variablesSetCmd.Flags().Bool("replace", false, "Fully replace previous variables instead of updating them")
```
JakeCooper, line 157:
```suggestion
	variablesSetCmd.Flags().Bool("replace", false, "Fully replace all previous variables instead of updating them")
```

main.go
```diff
@@ -154,6 +154,8 @@ func init() {
...
+	variablesSetCmd.Flags().Bool("yes", false, "Skip questions when replacing variables")
```
JakeCooper, line 158:
```suggestion
	variablesSetCmd.Flags().Bool("yes", false, "Skip all confirmation dialogs")
```

controller/envs.go
```diff
@@ -185,13 +185,13 @@ func (c *Controller) UpsertEnvs(ctx context.Context, envs *entity.Envs, serviceN
...
 
-	return c.gtwy.UpsertVariablesFromObject(ctx, &entity.UpdateEnvsRequest{
+	return c.gtwy.UpdateVariablesFromObject(ctx, &entity.UpdateEnvsRequest{
 		ProjectID:     projectCfg.Project,
 		EnvironmentID: projectCfg.Environment,
 		PluginID:      pluginID,
 		ServiceID:     serviceID,
 		Envs:          envs,
-	})
+	}, replace)
```
JakeCooper, line 194: Any shot we can put Replace as a property on the entity?

JakeCooper (approved):
Few comments but looks great overall! Added the release label for you

JakeCooper (approved):
Official stamp will merge when tests pass!

PR #267, opened 2022-08-21
Fix outdated `deployEnvironmentTriggers` mutation

### LINKED ISSUE #266: Redeployment with new variables from `railway variables set` fails with status code 400
At the moment updating variables with `railway variables set` is not working correctly. The variables get updated correctly, but the triggered redeploy fails with status code 400. My guess is the endpoint was slightly changed broken that part of the command. This problem has been reproduced by another member on the Discord server which indicates it's most likely wrong on your end.

image

ChristianIvicevic:
This draft attempts to fix an outdated mutation to trigger a redeploy. I attempted to reuse the inferred `serviceID` and return it to the original callsite causing some changes here and there.

Unfortunately during local testing the mutation was successfully invoked but no redeployment was initiated server-side. I guess this is just filtered based on the headers disallowing CLI redeployments which may be intended for now or just a bug. The very same mutation fired from the UI by updating any variable does work though. As such this probably requires somebody to check that on your end.

Another observation I made is the type of mutation parameters. The server returns a 400 status code when the types are declared as `String!` instead of `ID!`. This is highly inconsistent with other queries and mutations expecting `String!` instead. My assumption is that this is a very old endpoint you exposed and haven't updated it yet.

I'll keep this a draft until the open questions are resolved and the correctness of the implementation has been confirmed.

Closes #266

### DIFF
--- a/cmd/variables.go
+++ b/cmd/variables.go
@@ -148,7 +148,12 @@ func (h *Handler) redeployAfterVariablesChange(ctx context.Context, environment 
 	}
 
 	ui.StopSpinner("Deploy triggered")
-	// TODO: This link is outdated and requires the ID of the deployment in progress and the service ID. Maybe DeployEnvironmentTriggers could return it to build the correct URL.
-	fmt.Printf("☁️ Deploy Logs available at %s\n", ui.GrayText(h.ctrl.GetProjectDeploymentsURL(ctx, latestDeploy.ProjectID)))
+
+	deployment, err := h.ctrl.GetLatestDeploymentForEnvironment(ctx, latestDeploy.ProjectID, environment.Id)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("☁️ Deploy Logs available at %s\n", ui.GrayText(h.ctrl.GetServiceDeploymentsURL(ctx, latestDeploy.ProjectID, *serviceID, deployment.ID)))
 	return nil
 }
--- a/cmd/variables.go
+++ b/cmd/variables.go
@@ -57,10 +57,10 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest) 
 		return err
 	}
 
-	noRedeploy, err := req.Cmd.Flags().GetBool("no-redeploy")
+	skipRedeploy, err := req.Cmd.Flags().GetBool("skip-redeploy")
 	if err != nil {
 		// The flag is optional; default to false.
-		noRedeploy = false
+		skipRedeploy = false
 	}
 
 	variables := &entity.Envs{}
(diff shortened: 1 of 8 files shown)


ChristianIvicevic:
After some discussion on Discord the endpoint was fixed and I could confirm that the CLI command now works. I was also able to deduce the correct link for the new deployment log.

cmd/variables.go
```diff
@@ -72,7 +78,7 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest)
 		updatedEnvNames = append(updatedEnvNames, key)
 	}
 
-	err = h.ctrl.UpsertEnvs(ctx, variables, &serviceName)
+	serviceID, err := h.ctrl.UpsertEnvs(ctx, variables, &serviceName)
```
JakeCooper, line 81: Would rather have an extra call to get the serviceId from serviceName than to change the function signature here

cmd/variables.go
```diff
@@ -57,6 +57,12 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest)
 		return err
 	}
 
+	noRedeploy, err := req.Cmd.Flags().GetBool("no-redeploy")
```
JakeCooper, line 60: Just copy nit here
```suggestion
	noRedeploy, err := req.Cmd.Flags().GetBool("skip-redeploy")
```

cmd/variables.go
```diff
@@ -86,7 +92,7 @@ func (h *Handler) VariablesSet(ctx context.Context, req *entity.CommandRequest)
 	fmt.Print(ui.Heading(fmt.Sprintf("Updated %s for \"%s\"", strings.Join(updatedEnvNames, ", "), environment.Name)))
 	fmt.Print(ui.KeyValues(*variables))
 
-	err = h.redeployAfterVariablesChange(ctx, environment)
+	err = h.redeployAfterVariablesChange(ctx, environment, serviceID, noRedeploy)
 	if err != nil {
```
JakeCooper, line 96: Ideally the redeployAfterVariablesChange shouldn't be aware of the flag. Should just have an `if !noRedeploys` at the top level here and keep the function signature clean

cmd/variables.go
ChristianIvicevic: I have been thinking about something similar but went naively with that approach since `UpsertEnvs` does a lot of steps to obtain the serviceID which is why I wanted to reuse the value it just computed. Can you elaborate what alternative you had in mind?

JakeCooper: Hoist the work required to get serviceId from name into another function, call that function above, then pass in the serviceId if you want

ChristianIvicevic: I see. The same snippet to obtain the serviceID by name was actually copy & pasted multiple times across the codebase already. I'll have a look at it, but can't promise anything since it's also used in unrelated parts of the codebase.

JakeCooper: Oh you don't have to refactor all of it. I'd just refactor it inline here and then we can move to pulling the usage into one function at some point

ChristianIvicevic: Alright, I came with something, but wasn't fully sure where to put it exactly to not copy & paste even more. Hope that this is better.

JakeCooper (approved):
Awesome!

cmd/variables.go
JakeCooper: Looks perfect great job!

PR #270, opened 2022-09-08
feat: Take $XDG_CONFIG_HOME into account

### LINKED ISSUE #206: CLI should respect user‘s XDG base directory settings
Railway CLI creates a config directory by default in `$HOME`. While this is not an error, I think the CLI should respect an OS (or user’s) XDG base directory settings. The config directory should be placed under `XDG_CONFIG_HOME`, if set.

I’d be happy to make a PR if you guys agree with the change.

Guergeiro:
Closes #206


Guergeiro:
@JakeCooper is this a good solution for the problem?

PR #296, opened 2023-01-10
Fix evaluation of symlinks on up

gschier:
We were checking if the file was a directory before evaluating symlinks. This PR fixes that order to prevent treating linked directories as files.


JakeCooper (approved):
Amazing

PR #304, opened 2023-03-03
Migrate to CLI v3

nebulatgs:
The repository tracking changes can be found at https://github.com/railwayapp/cliv3

**Summary of Changes**
- Migrate to using Rust with Clap for CLI args
- Use the v2 Public API
- Support for `railway service` to link to a service
- `railway domain` to generate a service domain
- Add various flags for use in CI/CD scripts


.github/workflows/release.yml
```diff
@@ -0,0 +1,154 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      rlwy_version: ${{ env.CLI_VERSION }}
+
+    steps:
+      - name: Get the release version from the tag
+        shell: bash
+        if: env.CLI_VERSION == ''
+        run: |
+          # Apparently, this is the right way to get a tag name. Really?
```
ndneighbor, line 21: I ran into the same thing a while ago. Some data model stuff for GH Actions don't make sense.

.github/workflows/label-check.yml
```diff
@@ -11,6 +11,6 @@ jobs:
       - uses: jesusvasquez333/verify-pr-label-action@v1.4.0
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
-          valid-labels: "release/patch, release/minor, release/major"
-          pull-request-number: '${{ github.event.pull_request.number }}'
+          valid-labels: "release/patch, release/minor, release/major, release/skip"
```
ndneighbor, line 14: Good add! :+1:

.github/workflows/tag.yml
```diff
@@ -1,58 +0,0 @@
-name: Tag
```
ndneighbor, line 1: Slight nit: this comes from years of managing yml. I personally find .yml files unreadable whenever you have any yml file above 100 lines. There is some opportunity to maybe split the .ymls and steps, but I think this is fine for now and dependent on personal opinion.

Why do you think we should have merged the files in your opinion?

CONTRIBUTING.md
```diff
@@ -1,15 +1,17 @@
...
-Build
 
-```shell
-make build
-```
+### Nix Setup
+`nix-shell` to enter the shell with all the dependencies
+
+### Running and Building
+`cargo run -- <args>` to run the binary\
```
ndneighbor, line 16: Trailing slash here, I assume you fat fingered that key. 😃

Cargo.toml
```diff
@@ -0,0 +1,74 @@
+[package]
+name = "railwayapp"
+version = "0.0.15"
```
ndneighbor, line 3: I assume the version would have to be set here?

CONTRIBUTING.md
nebulatgs: The trailing slash indicates a newline in Markdown, so it's needed

src/commands/starship.rs
```diff
@@ -0,0 +1,13 @@
...
+
+/// Starship Metadata
+#[derive(Parser)]
+#[clap(hide = true)]
+pub struct Args {}
+
+pub async fn command(_args: Args, _json: bool) -> Result<()> {
+    let configs = Configs::new()?;
+    let linked_project = configs.get_linked_project().await?;
+    println!("{}", serde_json::to_string(&linked_project)?);
```
coffee-cup, line 11: I don't understand how to use this command. Maybe add some descriptive docs?

image

uninstall.sh
```diff
@@ -1,3 +0,0 @@
-rm $(which railway)
```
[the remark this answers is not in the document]
nebulatgs: The install script can be run with the `-r` flag to uninstall

src/commands/starship.rs
nebulatgs: This is a hidden command for `starship` (https://github.com/starship/starship)
It's used to get metadata for the prompt - it doesn't appear in help and is not meant for direct use by people.

Cargo.toml
nebulatgs: Cargo release should set this once we release it to crates.io

src/commands/domain.rs
```diff
@@ -0,0 +1,106 @@
...
+        bail!("No services found for project");
+    }
+
+    // If there is only one service, it will generate a domain for that service
+    let service = if body.project.services.edges.len() == 1 {
+        body.project.services.edges[0].node.clone().id
+    } else {
+        let Some(service) = linked_project.service.clone() else {
+        bail!("No service linked");
+      };
```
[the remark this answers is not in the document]
nebulatgs: Added a check for this

src/commands/domain.rs
```diff
@@ -0,0 +1,106 @@
...
+        post_graphql::<queries::ServiceDomains, _>(&client, configs.get_backboard(), vars).await?;
+
+    let body = res
+        .data
+        .context("Failed to retrieve to get domains for service.")?;
+
+    let domain = body.domains;
+    if domain.service_domains.is_empty() || domain.custom_domains.is_empty() {
+        bail!("Domain already exists on service");
+    }
```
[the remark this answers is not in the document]
nebulatgs: Fixed by negating

src/commands/add.rs
```diff
@@ -0,0 +1,128 @@
...
+        if filtered.is_empty() {
+            bail!("Plugins already exist");
+        }
+
+        filtered
+    } else {
+        inquire::MultiSelect::new("Select plugins to add", filtered_plugins)
+            .with_render_config(render_config)
+            .prompt()?
+    };
```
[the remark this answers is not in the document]
JakeCooper: When we can, please try to keep code linear

Logic trees should be avoided if early returns will do (which they usually will)

src/commands/domain.rs
```diff
@@ -0,0 +1,106 @@
...
+    let body = res.data.context("Failed to retrieve response body")?;
+
+    if body.project.services.edges.is_empty() {
+        bail!("No services found for project");
+    }
+
+    // If there is only one service, it will generate a domain for that service
+    let service = if body.project.services.edges.len() == 1 {
+        body.project.services.edges[0].node.clone().id
+    } else {
```
JakeCooper, line 36: Does line 35 return? If so, we should be using explicit returns

Implicit returns are fine for small functions like map/filter/return but should be avoided, for clarity, here

Also, if the else is going to bail, we should flip the branches and have an early return to remove the else case

src/config.rs
```diff
@@ -0,0 +1,279 @@
...
+        let path = self.get_closest_linked_project_directory()?;
+        let project = self
+            .root_config
+            .projects
+            .get(&path)
+            .context("Project not found! Run `railway link` to link to a project")?;
+        Ok(project.clone())
+    }
+
+    pub fn get_linked_project_mut(&mut self) -> Result<&mut RailwayProject> {
```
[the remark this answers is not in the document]
nebulatgs: This is for mutating the project specified in the configuration file, project tokens work without touching the config file, so we don't need to handle them here

src/consts.rs
```diff
@@ -0,0 +1,14 @@
...
+pub const TICK_STRING: &str = "⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏ ";
+
+pub const PLUGINS: &[&str] = &["PostgreSQL", "MySQL", "Redis", "MongoDB"];
+
+pub const NO_SERVICE_LINKED: &str =
+    "No service linked and no plugins found\nRun `railway service` to link a service";
+pub const ABORTED_BY_USER: &str = "Aborted by user";
+pub const PROJECT_NOT_FOUND: &str = "Project not found!";
+pub const SERVICE_NOT_FOUND: &str = "Service not found!";
+pub const NON_INTERACTIVE_FAILURE: &str = "This command is only available in interactive mode";
```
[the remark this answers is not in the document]
nebulatgs: Not sure, is that more idiomatic?

src/commands/link.rs
```diff
@@ -0,0 +1,270 @@
...
+                body.project
+                    .environments
+                    .edges
+                    .iter()
+                    .map(|env| ProjectEnvironment(&env.node))
+                    .collect(),
+            )
+            .with_render_config(configs.get_render_config())
+            .prompt()?
+        };
```
JakeCooper, line 70: Logic that probably would benefit from being pulled into a function

JakeCooper (commented):
I'd say in general we need to take a pass over this to make it more clear and clean

I know you're trying to land this for today, so we can just fix the flags I've added for now, but we're gonna want to go over this again and iron it out because IMO, there are a few places for which it's a bit awkward

src/consts.rs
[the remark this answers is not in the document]
coffee-cup: I think would be a little bit. But very low priority in the grand scheme of things

src/commands/add.rs
```diff
@@ -0,0 +1,128 @@
...
+                        .tick_chars(TICK_STRING)
+                        .template("{spinner:.green} {msg}")?,
+                )
+                .with_message(format!("Creating {plugin}..."));
+            spinner.enable_steady_tick(Duration::from_millis(100));
+            post_graphql::<mutations::PluginCreate, _>(&client, configs.get_backboard(), vars)
+                .await?;
+            spinner.finish_with_message(format!("Created {plugin}"));
+        }
+    }
```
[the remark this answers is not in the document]
coffee-cup: Or just abstract it ourselves. We shouldn't have to duplicate the gql query every time depending on if we are in a terminal environment or not

src/commands/add.rs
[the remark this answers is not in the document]
nebulatgs: This isn't an early return, it's a conditional value (similar to a ternary expression in other languages), so the else is needed

src/commands/domain.rs
nebulatgs: This another instance of a conditional value

src/commands/init.rs
```diff
@@ -0,0 +1,164 @@
...
+        .with_render_config(render_config)
+        .prompt()?;
+
+    let name = if name.is_empty() {
+        use names::Generator;
+        let mut generator = Generator::default();
+        generator.next().context("Failed to generate name")?
+    } else {
+        name
+    };
```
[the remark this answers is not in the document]
coffee-cup: Agreed. In general I think we should abstract out the `inquire` library so we can do things like

```rs
let name = prompt("Project Name", "{placeholder}", "{help message}").or_else(|| random_name())
```

or some variation of that. Basically condensing these lines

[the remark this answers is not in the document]
nebulatgs: I think this is a great idea, will do!

ndneighbor (approved):
Green, expected fast follow on for code quality.

PR #306, opened 2023-03-04
fix release workflow

nebulatgs:
Renames the binary from `rlwy` to `railway`


nebulatgs:
Also fixes the install script (points at the main repo)

ndneighbor (approved):

PR #307, opened 2023-03-04
fix<install>: changed URLs to point to new repo

alexng353:


ndneighbor:
We have a fix incoming- but great catch!

ndneighbor:
Fix live!

PR #334, opened 2023-03-06
allow user to select service on `up`

nebulatgs:
fix 409 errors on projects with mutiple services
fix non-interactive checks

usage
```bash
railway up --service <SERVICE_NAME> [PATH]
```
or
```bash
railway up --service <SERVICE_ID> [PATH]
```

### DIFF
--- a/src/commands/add.rs
+++ b/src/commands/add.rs
@@ -78,11 +78,7 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
             project_id: linked_project.project.clone(),
             name: plugin.to_lowercase(),
         };
-        if !std::io::stdout().is_terminal() {
-            println!("Creating {}...", plugin);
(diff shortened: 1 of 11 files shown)


src/commands/add.rs
```diff
@@ -94,6 +90,10 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
             post_graphql::<mutations::PluginCreate, _>(&client, configs.get_backboard(), vars)
                 .await?;
             spinner.finish_with_message(format!("Created {plugin}"));
+        } else {
+            println!("Creating {}...", plugin);
+            post_graphql::<mutations::PluginCreate, _>(&client, configs.get_backboard(), vars)
```
coffee-cup, line 95: Doesn't have to be addressed in this PR, but we should really find a way to combine the tty/non-tty flows. Having to call the mutation twice per command is not great at all.

JakeCooper (approved):

coffee-cup (approved):

PR #335, opened 2023-03-06
make script compatible with `dash`

nebulatgs:
This fixes the install script so it works in `dash`, the default `sh` on most Linux distros

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -55,7 +55,7 @@ info() {
 }
 
 debug() {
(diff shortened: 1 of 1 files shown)


gschier (approved):
Seems good, assuming it's been tested everywhere 👍🏻

nebulatgs:
CleanShot 2023-03-06 at 17 52 59@2x

PR #342, opened 2023-03-08
Render service variables for `railway variables`, `railway shell`, and `railway run`

nebulatgs:
Fixes the variable rendering on these commands, before it was very hacky and didn't fully render the vars.

### DIFF
--- a/src/commands/shell.rs
+++ b/src/commands/shell.rs
@@ -27,14 +27,6 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
     let mut all_variables = BTreeMap::<String, String>::new();
     all_variables.insert("IN_RAILWAY_SHELL".to_owned(), "true".to_owned());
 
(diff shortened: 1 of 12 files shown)


src/commands/shell.rs
```diff
@@ -27,34 +27,14 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
     let mut all_variables = BTreeMap::<String, String>::new();
     all_variables.insert("IN_RAILWAY_SHELL".to_owned(), "true".to_owned());
 
-    let plugins: Vec<_> = body
+    let _plugins: Vec<_> = body
```
JakeCooper, line 30: This is unused right? If so please delete it instead of burying the lint error

JakeCooper (approved):
Minor stylistic things I'm glossing over here so that we can get this unblocked. We'll come back to those later

PR #346, opened 2023-03-09
fix `run` command environment arg resolution

nebulatgs:
The environment argument was not being resolved, and was assumed to be an ID

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -1,15 +1,12 @@
-use std::collections::BTreeMap;
-
 use anyhow::bail;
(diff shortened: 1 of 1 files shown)


ndneighbor (approved):
General question: is arg parsing on a per command basis? Meaning, are we likely to run into issues with arg parsing on other commands?

nebulatgs:
> General question: is arg parsing on a per command basis? Meaning, are we likely to run into issues with arg parsing on other commands?

In this case, yes. I'll try to share args between certain commands in a future PR

PR #350, opened 2023-03-10
fix: npm install on windows

### LINKED ISSUE #348: NPM install still fails in Windows 11
Windows - 11
Node.js - v18.13.0
Terminal - Git Bash

image

aleksrutins:
This should fix #348, although I don't have a Windows to test it on.

This also adds `package-lock.json` to Git - if y'all don't want that, I can remove it.


moinulmoin:
I guess probably instead of .tar.gz, it should be .zip for windows?

PR #351, opened 2023-03-10
Add --environment flag for railway up

### LINKED ISSUE #331: BUG - New V3 doesn't support service parameter inside up command.
Hey, with new major version my gitlab CI starts crashing from yesterday.. I was able to partly migrate to v3, but i am stuck with `railway up` command. In v2 there was property to specify which service i wanna turn on  `--service`. In new v3 there is not such a option and i can't find anywhere how to specify that.
Is there any alternative for this option or you just forgot to implement this key feature in new version ?

aleksrutins:
Should fix #331 

I haven't tested it yet, though.


uragecz:
@aleksrutins  but this fix / adds only environemtn flag right ? But there is issue with --service flag as well

aleksrutins:
The `--service` flag already exists.

uragecz:
@aleksrutins Read that issue once again.  It's about `--service` flag isn't working. It keep saying` error: unexpected argument '--service' found`

coffee-cup:
I've updated this PR a bit to allow specifying the environment by name (or id)

PR #381, opened 2023-03-12
fix generation of `.tar.gz` archives on windows

### LINKED ISSUE #348: NPM install still fails in Windows 11

devtomio:
maybe fixes #348


maddsua (approved):
Cant' test it now, but to me your solution seems the most appropriate

moinulmoin (approved):
It should be good to go.

maddsua:
Yep, should be good. Specifying `... filename.tar.gz ...` on a single input does not make 7z create tar.gz, it just does gz instead

nebulatgs (approved):

PR #388, opened 2023-03-18
Ignore CTRL+C on `railway run`

### LINKED ISSUE #384: CTRL + C drops users out of non-railway shells
Flow: 

* A user starts a `rails c` console session
* A user runs `railway run`
* A user hits CTRL + C

Expected Behavior:
* The user is dropped out of the `railway run` command but NOT the surrounding rails console

Observed Behavior:
* The user is dropped out of the `railway run` command AND the surrounding rails console

Suggested Resolution:
* Capture the CTRL + C and do not propagate it upwards

Workaround:
* Use `railway shell` and then use `rails c`

nebulatgs:
This was causing issues where both the child and parent processes would exit at the same time, and the child process would be unable to respond to the SIGINT.

This PR ignores SIGINT on the parent process, and allows the child to handle it.

Resolves #384

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -352,6 +352,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctrlc"
+version = "3.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbcf33c2a618cbe41ee43ae6e9f2e48368cd9f9db2896f10167d8d762679f639"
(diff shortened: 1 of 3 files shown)


gschier (approved):

nebulatgs:
Something funny is happening with this PR, it seems to include a commit already on `master`

JakeCooper:
Try merge/rebase?

PR #392, opened 2023-03-22
Update `open` off of yanked version

dbanty:
Currently, you have to `cargo install --locked`, and you get a warning about the yanked version. It seems like the only breaking change to v4 is upping the MSRV (still lower than what's declared for this project).


gschier (approved):
Nice! Tested and it seems to work 👍🏻

PR #394, opened 2023-03-22
Fix config writes

### LINKED ISSUE #385: unable to parse config file, regenerating
happens when attempting to use the v3 cli with a config that was generated with the v2 cli

gschier:
Fixes #385

This PR updates how the CLI writes configs. It now write atomically, by first writing to a tmp file, then renaming to the final destination.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,17 +1,13 @@
-use std::{
-    collections::BTreeMap,
-    fs::{create_dir_all, File},
-    io::{Read, Write},
-    path::PathBuf,
-};
+use std::{collections::BTreeMap, fs::{create_dir_all, File}, fs, io::Read, path::PathBuf};
+use std::env::temp_dir;
 
 use anyhow::{Context, Result};
 use colored::Colorize;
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -270,15 +266,22 @@ impl Configs {
...
-        };
-        let serialized_config = serde_json::to_vec_pretty(&self.root_config)?;
-        file.write_all(serialized_config.as_slice())?;
-        file.sync_all()?;
+
+        // Create temporary file to write initial data to. This is to ensure updates
+        // are atomic. After writingn the tmp file, we will rename it to the final destination,
+        // which is an atomic operation.
+        let mut tmp_file_path = temp_dir();
+        tmp_file_path.push(self.root_config_path.file_name().unwrap());
```
nebulatgs, line 276: I think we should avoid `.unwrap()`, you should be able to attach `.context()` to it and then `?` to return it as a `Result::Err()`.

nebulatgs (approved):

PR #409, opened 2023-04-20
feat: windows specific shell detection

alexng353:
~~Uses [wmic](https://stackoverflow.com/questions/7486717/finding-parent-process-id-on-windows) and `std::process::id` to detect the parent process and therefore the correct shell to drop the user into.~~

Uses winapi syscalls and the `std::process::id` to detect the parent process of the cli. This either detects a shell listed below or defaults to `cmd`. Does nothing on non-windows platforms because of `#[cfg(target_os = "windows")]`

Supported shells:
- cmd
- Powershell 5
- Powershell 7 (pwsh)
- NuShell
- ElvSh

Edits:
- Because of brody, this now implements [this gist](https://gist.github.com/mattn/253013/d47b90159cf8ffa4d92448614b748aa1d235ebe4)

### DIFF
--- a/src/commands/shell.rs
+++ b/src/commands/shell.rs
@@ -102,7 +102,6 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
     }
 
     /// https://gist.github.com/mattn/253013/d47b90159cf8ffa4d92448614b748aa1d235ebe4
-    /// Only matches cmd, powershell or pwsh for safety
     /// defaults to cmd if no parent process is found
     #[cfg(target_os = "windows")]
     async fn windows_shell_detection() -> Option<WindowsShell> {
--- a/src/commands/shell.rs
+++ b/src/commands/shell.rs
@@ -108,13 +108,23 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
     /// defaults to cmd if no parent process is found
     #[cfg(target_os = "windows")]
     async fn windows_shell_detection() -> Option<WindowsShell> {
-        let (_, ppname) = unsafe {
+        let (ppid, mut ppname) = unsafe {
             get_parent_process_info()
                 .context("Failed to get parent process info")
                 .unwrap_or_else(|_| (0, "".to_string()))
         };
 
+        if ppname == "node.exe" {
+            (_, ppname) = unsafe {
+                node_fix(ppid)
+                    .context("Failed to get parent process info")
+                    .unwrap_or_else(|_| (0, "".to_string()))
+            }
+        }
+
         let ppname = ppname.split(".").next().unwrap_or("cmd");
+
+        dbg!(ppname);
 
         match ppname {
             "cmd" => Some(WindowsShell::Cmd),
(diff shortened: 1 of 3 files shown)


src/commands/shell.rs
```diff
@@ -73,20 +93,135 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
+    // disable dead code warning for windows_shell_detection
+    #[allow(dead_code)]
+    enum WindowsShell {
+        Cmd,
+        Powershell,
+        Powershell7,
+    }
+
+    /// https://gist.github.com/mattn/253013/d47b90159cf8ffa4d92448614b748aa1d235ebe4
+    /// Only matches cmd, powershell or pwsh for safety
```
aleksrutins, line 105: Consider adding:
- nushell
- elv.sh

src/commands/shell.rs
```diff
@@ -73,20 +93,135 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
         .context("Failed to wait for command")?;
+
     println!("Exited subshell, Railway variables no longer available.");
     Ok(())
 }
+
+/// get the parent process info, translated from
+// https://gist.github.com/mattn/253013/d47b90159cf8ffa4d92448614b748aa1d235ebe4
+#[cfg(target_os = "windows")]
+fn get_parent_process_info() -> Option<(DWORD, String)> {
```
aleksrutins, line 162: This whole function should probably be marked as unsafe.

src/commands/shell.rs
```diff
@@ -73,20 +93,135 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
+
+    if let Some(ppname) = parent_process_name {
+        Some((ppid, ppname))
+    } else {
+        None
+    }
+}
+
+#[cfg(target_os = "windows")]
+fn get_process_name(pid: DWORD) -> Option<String> {
```
aleksrutins, line 201: unsafe, etc.

aleksrutins (approved):
LGTM, just a couple suggestions.

src/commands/shell.rs
alexng353: Unable to test until later tonight. Will commit then.

Thanks for the feedback.

half0wl (commented):
Nice @alexng353 ❤️‍🔥 

Quick thoughts: Would this detection be possible without FFIs? I imagine it's not as robust, but unsafe code tends to end up being a foot gun somewhere. Or perhaps we could replace the unsafe calls with the [`winsafe`](https://docs.rs/winsafe/latest/winsafe/index.html) crate? From a very quick skim, it looks like all the APIs this is relying on is available there.

Tagging in @coffee-cup for a detailed review

alexng353:
Regarding the FFIs, I don't believe it's possible without either a call to a winapi or WMIC. I'm not too sure about the winsafe crate but I'll take a look at it tonight. Thanks.

src/commands/shell.rs
alexng353: Added nushell and elv.sh

src/commands/shell.rs
alexng353: Marked as unsafe

src/commands/shell.rs
alexng353: Marked as unsafe

alexng353:
Decided not to rewrite it in winsafe because it's not really worth it. 🙂✌️

src/commands/shell.rs
```diff
@@ -73,20 +91,178 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
+    enum WindowsShell {
+        Cmd,
+        Powershell,
+        Powershell7,
+        NuShell,
+        ElvSh,
+    }
+
+    /// https://gist.github.com/mattn/253013/d47b90159cf8ffa4d92448614b748aa1d235ebe4
+    /// Only matches cmd, powershell or pwsh for safety
```
coffee-cup, line 105:
```suggestion
```

src/commands/shell.rs
```diff
@@ -73,20 +91,178 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
         eprintln!("No service linked, skipping service variables");
     }
 
+    // disable dead code warning for windows_shell_detection
+    #[allow(dead_code)]
+    enum WindowsShell {
```
coffee-cup, line 96: Can this and all shell detection code be extracted outside of the `command` function

alexng353: Will do

coffee-cup (approved):

coffee-cup:
Thanks! 🎉

PR #410, opened 2023-04-21
fix<run>: allows `railway run` to run batch scripts on windows, including pnpm and npm.

### LINKED ISSUE #386: `railway run` can't run batch files
2023-03-16 20_52_47-dayjs

this is more so of an issue since `npx` is a cmd file on windows, thus you can't do `railway run npx ...`

the workaround is `railway shell` and then run the `npx` commands (or batch files)

alexng353:
Fixes #386
Allows `railway run` to run other batch script windows programs, such as pnpm and npm.

- one match statement that sets the child proc and does all the argument handling
- moved command spawning away from the match statemnt
- handles no argument passed

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -144,13 +144,28 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
         // this is for `rails c` and similar REPLs
     })?;
 
-    let exit_status =
-        tokio::process::Command::new(args.args.first().context("No command provided")?)
-            .args(args.args[1..].iter())
-            .envs(variables)
-            .status()
-            .await
-            .context("Failed to spawn command")?;
+    let exit_status: std::process::ExitStatus;
+
+    match std::env::consts::OS {
+        "windows" => {
+            exit_status = tokio::process::Command::new("cmd")
+                .arg("/C")
+                .args(args.args.iter())
+                .envs(variables)
+                .status()
+                .await
+                .context("Failed to spawn command")?;
+        }
+        _ => {
+            exit_status =
+                tokio::process::Command::new(args.args.first().context("No command provided")?)
+                    .args(args.args[1..].iter())
+                    .envs(variables)
+                    .status()
+                    .await
+                    .context("Failed to spawn command")?;
+        }
+    }
 
     if exit_status.success() {
         println!("Looking good? Run `railway up` to deploy your changes!");
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -144,8 +144,7 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
         // this is for `rails c` and similar REPLs
     })?;
 
-    let slash_c = "/C".to_owned();
-    let mut args = args.args.iter().collect::<Vec<_>>();
+    let mut args = args.args.iter().map(|s| s.as_str()).collect::<Vec<_>>();
     if args.is_empty() {
         bail!("No command provided");
     }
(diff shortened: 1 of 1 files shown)


brody192:
i like

src/commands/run.rs
```diff
@@ -144,13 +144,30 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
     })?;
 
-    let exit_status =
-        tokio::process::Command::new(args.args.first().context("No command provided")?)
-            .args(args.args[1..].iter())
-            .envs(variables)
-            .status()
-            .await
-            .context("Failed to spawn command")?;
+    let slash_c = "/C".to_owned();
```
aleksrutins, line 147: There _has_ to be a better way to do this...

Down below, you could probably replace `&slash_c` with `"/C".into()`, right?

aleksrutins (approved):
LGTM, just a bit of feedback.

src/commands/run.rs
alexng353: I'll make the change during lunch break, I'm not so familiar with rust stuff yet. Thanks for the comment!

alexng353: Can't borrow the "/C".into() value since it is dropped too quickly. Instead, I just mapped args.args into a Vec<&str>.

src/commands/run.rs
```diff
@@ -144,13 +144,30 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
+    let mut args = args.args.iter().collect::<Vec<_>>();
+    if args.is_empty() {
+        bail!("No command provided");
+    }
+
+    let child_process_name: &str;
```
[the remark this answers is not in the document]
alexng353: I just made the inter happy instead.

src/commands/run.rs
```diff
@@ -144,13 +144,28 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
-            .envs(variables)
-            .status()
-            .await
-            .context("Failed to spawn command")?;
+    let mut args = args.args.iter().map(|s| s.as_str()).collect::<Vec<_>>();
+    if args.is_empty() {
+        bail!("No command provided");
+    }
+
+    // So the linter doesn't get mad. I think this is pretty ugly.
```
coffee-cup, line 152:
```suggestion
```

src/commands/run.rs
```diff
@@ -144,13 +144,28 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
+    let child_process_name = match std::env::consts::OS {
+        "windows" => {
+            args.insert(0, "/C");
+            "cmd"
+        }
+        _ => {
+            args.remove(0);
```
coffee-cup, line 159: This doesn't seem to work

image

src/commands/run.rs
```diff
@@ -144,13 +144,28 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
+            args.first().context("No command provided")?
```
coffee-cup, line 160: We are removing the first arg from the args array but then want to return it from the match. This can be changed to just remove the first item and return it

```suggestion
            args.remove(0)
```

src/commands/run.rs
coffee-cup: If you want to remove the match, you could change to an if

```rust
let child_process_name = if std::env::consts::OS == "windows" {
  args.insert(0, "/C");
  "cmd"
} else {
  args.remove(0)
}
```

alexng353:
In the middle of checking some compatibility issues on non-windows brought up by @brody192

alexng353:
> In the middle of checking some compatibility issues on non-windows brought up by @brody192

I think that the issue I was chasing was fixed by @coffee-cup 's suggestion here. Couldn't replicate the bug on my mac or linux machine, so I think its good.
image

coffee-cup (approved):

PR #413, opened 2023-04-24
[command] [up] ignore .git and node_modules by default

alexng353:
I've written some code that ignores .git and node_modules by default. I've also added a small amount of documentation to the block that handles the actual compression of the gzipped tarball, since it is, in my opinion, quite hard to read.

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -127,6 +127,9 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
         let mut builder = WalkBuilder::new(path);
         builder.add_custom_ignore_filename(".railwayignore");
         builder.add_custom_ignore_filename(".gitignore");
+
+        builder.add_ignore(".git/");
+
         let walker = builder.follow_links(true).hidden(false);
         let walked = walker.build().collect::<Vec<_>>();
         if let Some(spinner) = spinner {
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -142,10 +142,8 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
 
     // list of all paths to ignore by default
     let ignore_paths = vec![".git", "node_modules"];
-    let ignore_paths: Vec<&std::ffi::OsStr> = ignore_paths
-        .iter()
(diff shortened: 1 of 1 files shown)


src/commands/up.rs
```diff
@@ -117,22 +117,49 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
         builder.add_custom_ignore_filename(".railwayignore");
         builder.add_custom_ignore_filename(".gitignore");
+
         let walker = builder.follow_links(true).hidden(false);
         let walked = walker.build().collect::<Vec<_>>();
         if let Some(spinner) = spinner {
             spinner.finish_with_message("Indexed");
         }
         if std::io::stdout().is_terminal() {
+            //#region
```
coffee-cup, line 162: What is this for?

src/commands/up.rs
```diff
@@ -143,17 +170,30 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
                 .with_finish(ProgressFinish::WithMessage("Compressed".into()));
             pg.enable_steady_tick(Duration::from_millis(100));
+            //#endregion
 
             for entry in walked.into_iter().progress_with(pg) {
                 let entry = entry?;
                 let path = entry.path();
+                if path
+                    .components()
+                    .any(|c| ignore_paths.contains(&c.as_os_str()))
```
coffee-cup, line 180: Wondering if this should be `starts_with` instead?

alexng353: It's like that because I wanted to cover cases where there is git repo inside git repo. or in a monorepo where there is node_modules in a sub folder.

src/commands/up.rs
alexng353: An extension I was using to collapse the #region. I'll remove it on a quick push during lunch. Sorry

src/commands/up.rs
coffee-cup: Sounds good!

coffee-cup (approved):
Looks good pending lint fixes and removal of `#region`/`#endregion` comments

PR #414, opened 2023-04-25
Better CLI errors

coffee-cup:
This PR
- Uses custom RailwayErrors to cleanup error handling in many places. Eventually, we should be using these custom errors everywhere. For now, we just use them for a few critical paths
- Wrap all GraphQL errors. This allows us to show custom help messages when not authorized. It also cleans up the data fetching
- Abstract fetching of projects to a controller function. This is just the start of the data fetching refactor. Moving forward we should be using a Railway client for all Railway interaction


alexng353:
@brody192 it's happening!

PR #418, opened 2023-04-26
Stream both build and deploy logs for up command

coffee-cup:
This PR
- Streams both build and deploy logs when running the up command. This means that the logs will automatically show the correct output. I thought about switching the stream over, but was fighting too much with the Rust compiler and accessing the stream from multiple threads. Streaming both build and deploy logs should have low overhead so I don't think it is a problem.
- Stop streaming the logs and exit with error code 1 if the deployment fails
- Move deployment logs subscription fetching into controller functions
- New GQL query to fetch a single deployment

Fixes https://linear.app/railway/issue/PRO-506/railway-up-should-switch-to-tailing-application-logs-on-success

### DIFF
--- a/src/commands/logs.rs
+++ b/src/commands/logs.rs
@@ -1,8 +1,6 @@
-use futures::StreamExt;
+use crate::controllers::deployment::{stream_build_logs, stream_deploy_logs};
 
-use crate::subscription::subscribe_graphql;
-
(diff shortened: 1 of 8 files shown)


gschier (approved):
Ooo, nice solution streaming both at once!

PR #422, opened 2023-04-29
fix<shell>: add "railway.exe" to recursive check list for `scoop` ins…

alexng353:
…tallations

Scoop uses a wrapper executable named "railway.exe" to start the actual railway executable.


half0wl:
Does this fix #424?

alexng353:
> Does this fix #424?

I think it fixes #424 but its been a long time since I wrote it. Sorry for the slow reply.

PR #427, opened 2023-05-26
use `DATABASE_URL` for psql

brody192:
Currently `railway connect Postgres` asks the user for the database password, this issue isn't present with the other database types when using `connect` as they either use a URI or have the password passed in as a flag.

This PR uses the `DATABASE_URL` variable as the only argument to `psql`
Specifying a connection string has been supported since [psql 9.2.0](https://www.postgresql.org/docs/release/9.2.0/#AEN114358:~:text=option%20for%20views-,Allow%20libpq%20connection%20strings%20to%20have%20the%20format%20of%20a%20URI,-Add%20a%20single)


brody192:
[reason for this pr](https://discord.com/channels/713503345364697088/1111313066764800062)

coffee-cup (approved):

PR #455, opened 2023-10-06
Add attribute logging

Milo123459:
This adds support for attribute logging.

Screenshots:
image


src/util/logs.rs
```diff
@@ -0,0 +1,64 @@
...
+            if !attr.value.ends_with('"') {
+                attr.value.push('"');
+            }
+        }
+        // get attributes using a match
+        for attr in &log.attributes {
+            match attr.key.to_lowercase().as_str() {
+                "timestamp" | "ts" | "time" => {
+                    timestamp = Some(attr.value.clone().replace('"', ""))
+                }
```
gschier, line 24: I would prefer to use `log.timestamp` instead of the one from the attributes, to be consistent with dashboard behavior

src/util/logs.rs
```diff
@@ -0,0 +1,64 @@
+use crate::subscriptions;
+use colored::Colorize;
+
+pub fn format_attr_log(mut log: subscriptions::deployment_logs::LogFields) {
+    if !log.attributes.is_empty() {
+        let mut timestamp: Option<String> = None;
+        let mut level: Option<String> = None;
+        let message = log.message;
+        let mut others = Vec::new();
+        // for some reason, not all have "" around the value
```
gschier, line 10: Attributes are JSON, so `number` `null` `boolean` won't have quotes.

src/util/logs.rs
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+        // get attributes using a match
+        for attr in &log.attributes {
+            match attr.key.to_lowercase().as_str() {
+                "timestamp" | "ts" | "time" => {
+                    timestamp = Some(attr.value.clone().replace('"', ""))
+                }
+                "level" | "lvl" => level = Some(attr.value.clone()),
+                _ => others.push(format!(
```
gschier, line 26: There might be nested attributes in here too. I feel like we should probably just throw these away for now? Or perhaps cap attributes to a certain length?

Here's an example response with some nested stuff.

CleanShot 2023-11-17 at 14 16 12@2x

src/util/logs.rs
```diff
@@ -0,0 +1,64 @@
...
+            "{}={} {}={} {}={}{}{5} {}",
+            "timestamp".bright_cyan(),
+            timestamp.unwrap_or_default().purple(),
+            "level".bright_cyan(),
+            level.unwrap_or_default(),
+            "msg".bright_cyan(),
+            "\"".dimmed(),
+            message,
+            others.join(" ")
+        );
```
gschier, line 60: Thoughts on cleaning up this formatting a bit? I have a few thoughts:

- We probably don't need the labels for timestamp and level, since they're obvious on their own
- Message could be between level and attrs, differentiated by color
- Can we try parsing/reformatting the time to be something more readable? Similar to the UI like `Nov 17 13:55:34.123` would be great
- Level colored to match the severity (debug=gray, info=cyan, warn=yellow, error=red) then have the attr names be magenta

A good format (imo) would be something like:

CleanShot 2023-11-17 at 14 26 19@2x

src/util/logs.rs
```diff
@@ -0,0 +1,64 @@
...
+    } else {
+        println!("{}", log.message);
```
gschier, line 62: We could also use `log.timestamp` and `log.severity` to format these in a nice way (even for non-structured logs) like we do in the dashboard

gschier (changes requested):
A few comments, mostly on style/formatting. Really excited to get this in, though. Thanks for taking it on! :)

src/util/logs.rs
Milo123459: i think it looked better when they all had it - cc @brody192

src/util/logs.rs
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+        // get the level and colour it
+        let level = level.map(|level| {
+            // make it uppercase so we dont have to make another variable
+            // for some reason, .uppercase() removes formatting
+            let level = level.replace('"', "").to_uppercase();
+            match level.to_lowercase().as_str() {
+                "info" => level.blue(),
+                "error" => level.red(),
```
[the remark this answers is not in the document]
gschier: We should also make all the levels the same length (DBUG, INFO, WARN, ERRO)

src/util/logs.rs
```diff
@@ -0,0 +1,64 @@
...
+                attr.value.push('"');
+            }
+        }
+        // get attributes using a match
+        for attr in &log.attributes {
+            match attr.key.to_lowercase().as_str() {
+                "timestamp" | "ts" | "time" => {
+                    timestamp = Some(attr.value.clone().replace('"', ""))
+                }
+                "level" | "lvl" => level = Some(attr.value.clone()),
```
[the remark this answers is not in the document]
Milo123459: yeah, it is

src/util/logs.rs
Milo123459: if a custom timestamp is provided, should i just skip over it?

src/util/logs.rs
Milo123459: severity isn't a field on the log, do you mean the attributes? but that won't work as this would be for non attribute logs.

src/commands/logs.rs
```diff
@@ -108,14 +111,7 @@ pub async fn command(args: Args, json: bool) -> Result<()> {
         })
         .await?;
     } else {
-        stream_deploy_logs(deployment.id.clone(), |log| {
-            if json {
-                println!("{}", serde_json::to_string(&log).unwrap());
```
JakeCooper, line 113: I think this removes the ability to json format this? Unless I'm mistaken

Milo123459: nevermind, working on it

gschier (approved):
My apologies for taking so long on this one! I went on parental leave shortly after my initial review.

Milo123459:
JSON logging currently has a bug where every value has escaped strings at the front

Milo123459:
image

JSON logging know looks like this.

PR #456, opened 2023-10-10
Fix up from automatically selecting a service

Milo123459:
Previously, `up` and `logs` would pick a service if only one service was in the linked project. This often caused unwanted behaviour, so this PR changes it to a prompt.

Relevant thread: https://discord.com/channels/713503345364697088/1160376027365785711

### DIFF
--- a/src/commands/logs.rs
+++ b/src/commands/logs.rs
@@ -90,9 +90,7 @@ pub async fn command(args: Args, json: bool) -> Result<()> {
     if let Some(deployment_id) = args.deployment_id {
         deployment = deployments
             .iter()
-            .find(|deployment| {
-                deployment.id == deployment_id
-            })
+            .find(|deployment| deployment.id == deployment_id)
             .context("Deployment id does not exist")?;
     } else {
         // get the latest deloyment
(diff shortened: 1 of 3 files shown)


coffee-cup (approved):
Good call

GrantBirki:
@coffee-cup + @Milo123459 just something to bring to your attention in regards to this being a `patch` change. This change actually broke all of our deployments. Our CI systems were not passing in the `--service` flag and our projects only had one service in each. For our production deployments, we actually have to commit changes to `main` for our CI system to pick these up. This lead to deployments failing with the following error message:

```text
Multiple services found. Please specify a service to deploy to via the --service flag.
```

By removing the `if services.len() == 1` statement in this pull request, you effectively made the `--service` flag a required argument when using `railway up` and this should be considered a breaking change as existing commands / deployments would now fail without this parameter.

My ask is that in the future, a little more thought is put into how these changes my impact CI systems and not just users running these commands. Broken build / deployment systems is something no one wants when bumping a dependency by a single patch level.

> I still love railway btw but I just wanted to provide some constructive feedback about this issue that my team faced

Thanks! 🙇 ❤️

PR #463, opened 2023-10-25
Make `status` show linked service instead of all

Milo123459:
Instead of showing all services, it now shows just the linked one.


zuchka:
LGTM!

CleanShot 2023-11-17 at 12 10 54@2x

zuchka (approved):
LGTM! works as expected

PR #464, opened 2023-10-26
Improve login

Milo123459:
If you answer no to the prompt about opening the browser, instead of just exiting, it'll now perform a browserless login instead.


zuchka:
@Milo123459 can you confirm that this is how your feature should work? I got this prompt and then I clicked out to the browser link. Just want to double check.

CleanShot 2023-11-17 at 12 19 11@2x

Milo123459:
the wording is slightly confusing which could be changed in a follow up pr, but yes this is intended behaviour @zuchka

zuchka (approved):
🚀

Issue #469, opened 2023-11-27
[Feature request] Docker image for CI/CD


maddsua:
Since Railway does not have a direct integration with GitLab the recommended way of deploying stuff from GitLab is to spin up an empty debian container, install railway cli in it and run it from there. This method certainty works, but having a proper docker image would speed things up a lot. Also, alpine-based images tend to get ready noticeably faster than debian ones.

maddsua:
Closing this issues as the linked (not really linked but who cares) PR was merged

mattrossman:
@maddsua Can you elaborate how we'd use the image in that PR to fix this issue?

I can't tell if this image has been deployed anywhere for users or if we have to build and host it ourselves.

This article https://blog.railway.app/p/gitlab-ci-cd still mentions pulling the `ubuntu` image and installing Railway CLI into it, can you show an example GitLab pipeline configuration that uses the image with Railway preinstalled?

maddsua:
Hey @mattrossman

As of today, the image is still not deployed to ghcr (or any other registry for that matter). The railway team have not yet enabled containers feature in the repo for some reason. Also, the pipeline isn't even called anywhere at the moment. In other words, altho the dockerfile was merged, it really didn't get anywhere.

Speaking of the article, it's a blog post and I'm not expecting them to update it anytime soon.

About a pipeline example, I guess you wanted to see an example from where you can pull that image. Ummm yeah, you can't really do that at the moment.

My problem with their method was mainly caused by debian not having curl pre installed (so is alpine iirc), and having to run extra commands on each build wasn't something I wanted to do.

Currently my solution is to literally download a specific binary from their release page via wget and copy it directly to `/bin`. Not a pretty one, but at least it don't generate a wall full of install logs.

You definitely can publish that image yourself if you want to. Or actually, since it's bed time for me right now and I'm nowhere near being asleep I could publish it myself just for the sake of it.

maddsua:
Here it is lol
https://github.com/maddsua/freight/pkgs/container/freight

PR #470, opened 2023-11-27
feat: alpine dockerfile

alexng353:
Dockerfile for #469, not sure if I should do Dockerfile's for other distros or not. 

Also, I haven't made an actions workflow that automatically builds and publishes these yet. I assume that it would be based off of the [cargo-release](https://github.com/railwayapp/cli/blob/master/.github/workflows/cargo-release.yml) file, but I'm not sure.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+Dockerfile
+bin/
+npm-install/
+target/
+.cargo/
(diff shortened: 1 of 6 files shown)


coffee-cup (approved):
Thank you for this!

PR #477, opened 2024-01-21
Support CICD

maddsua:
### Changes:

- Added new `up` command argument: `cicd`
    When this argument is provided, cli won't stream deploy logs and will exit immediately after deploy status changes to `SUCCESS`.
    This required changing `wait_for_failure` function to support this case. In both of them it will cause the cli to exit with either error exit code or zero, which would indicate success.

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -66,8 +66,8 @@ pub struct UpErrorResponse {
 }
 
 enum UpExitReason {
-    DEPLOYED = 0,
-    FAILED = 1,
+    Deployed,
+    Failed,
 }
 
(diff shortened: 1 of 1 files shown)


maddsua:
By the way, can we make a release after this PR is merged? We had docker image ready for almost 3 months now and it's still not published

aleksrutins (approved):

coffee-cup (approved):
Thanks for this PR! And sorry it has taken so long to move forward. I've renamed `--cicd` to `--ci`, but everything else looks good. I will cut a release once everything is merged and green.

PR #478, opened 2024-01-26
feat: railway zsh plugin

iloveitaly:


railway.plugin.zsh
coffee-cup, line 1: I am not familiar with `zinit`. Will this file be pulled automatically?

coffee-cup (commented):
Thanks for the PR! Just a comment about the `.zsh` file

railway.plugin.zsh
iloveitaly: yes, if the `zinit load` line is used

yliyang (approved):

alexng353:
I'm pretty sure [zinit](https://github.com/zdharma-continuum/zinit) is not installed by default with zsh.

PR #481, opened 2024-02-15
Remove .gitignore from ignore files in `up`

Milo123459:
This removes the searching of .gitignore in the up command. This previously caused some unknown behaviour.

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -173,7 +173,6 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
         let mut archive = Builder::new(&mut parz);
         let mut builder = WalkBuilder::new(path);
         builder.add_custom_ignore_filename(".railwayignore");
-        builder.add_custom_ignore_filename(".gitignore");
 
         let walker = builder.follow_links(true).hidden(false);
         let walked = walker.build().collect::<Vec<_>>();


zuchka (approved):
lgtm!

brody192:
I can see this causing far more issues, including but not limited to a whole bunch of permission-denied errors during build from having the node_modules folder make it into the build process. Please keep in mind `railway up` is not used only from a github action where the files have already been ignored

aleksrutins:
I agree with Brody, this will break anybody's builds who's trying to use `railway up` to deploy a Node.js app from a Windows machine.

Milo123459:
> I agree with Brody, this will break anybody's builds who's trying to use `railway up` to deploy a Node.js app from a Windows machine.

why? `node_modules` is ignored by default.

maddsua:
This is what having .gitignore included makes me do in my workflows. Just gonna leave it here.
pic

brody192:
I'm in full agreeance with aleks, changing the default behavior that has been around since v1 of the cli is not the best idea, node_modules may be hard coded to be ignored, but this change **will** break other things in other unimaginable and subtle ways, for the people who need the cli to ignore the .gitignore, a simple flag like aleks suggested would be a perfect solution.

aleksrutins:
Also, in order to include specific files, (as I implemented it in v2) you can add a `.railwayignore` with `!` lines. Not 100% sure if that works in v3, though.

brody192:
I think the idea could work, but the `.gitignore` file is last to be added in v3, unlike in v2 where `.railwayignore` came last

maddsua:
> `deno task build` should be something you run in the Railway builder, no? The recommended workflow (as I understand it) is just to check if you can deploy it in your CI file, and then Railway's CI does the actual building.

Not in this case, railway wouldn't be able to build this project because of the way it's structured (a monorepo that is not approved by the lizard govt). My point isn't in about the build step but in the fact that I need to move that service to a temp directory which is not covered by gitignore in order for railway to be able to see all the needed files

maddsua:
Well if we can't change the default behavior, why not just add an option to disable gitignore? Something like `--no-gitignore` wouldn't hurt anybody

brody192:
I am in favor of leaving the default behavior and adding a `--no-gitignore` flag

Milo123459:
#484 has been made

PR #487, opened 2024-03-29
Remove references to plugins

### LINKED ISSUE #480: CLI still using legacy plugin-related commands / terminology
As you can see in the image, I need to do the command: railway add -p mysql to add the MySQL plugin to my project.
However, it asks to log in, says that I am not actorized, even though I have already logged in before.
I repeat Login and still receive the unauthorized message.
I tried other alternative commands for railway add, but to no avail.
(I don't know which data is really sensitive so I hid what I thought was best hidden.)

The image below has more details of what I tried to do.

I'm in the Ubuntu environment with WSL on Windows
latest version of railway installed

Capturar03

Milo123459:
Fixes #480
This PR:
- Removes the `delete` command
- Fixes the `add` command, allowing it to now deploy template powered databases
- Removes all references of plugins


zuchka (changes requested):
Milo,

This is looking fantastic. Can we please also update the docs here:
- https://github.com/railwayapp/docs/blob/main/src/docs/reference/cli-api.md

Ideally these should make no mention of plugins anymore. Do you agree?

Milo123459:
https://github.com/railwayapp/docs/pull/470 :)

zuchka (approved):

PR #489, opened 2024-04-12
Fix CI

Milo123459:
- **fix regular ci, does it work**
- **better audit ci**
- **change release ci**
- **fix release ci**

this in general just makes CI better, and automatically publishes to crates.io


zuchka (approved):
awesome!

.github/workflows/release.yml
```diff
@@ -167,3 +158,18 @@ jobs:
           download-url: https://github.com/railwayapp/cli/releases/latest/download/railway-${{ needs.create-release.outputs.railway_version }}-${{ matrix.target }}.tar.gz
         env:
           COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
+  publish-crate:
```
coffee-cup, line 161: I don't think this is needed. We publish with `cargo release` which automatically publishes to crates.io

coffee-cup: Unless the intention is to change the release process and initiate it via GitHub?

Milo123459: yes it is, having it run locally imo is a bad idea

.github/workflows/ci.yml
```diff
@@ -9,80 +9,25 @@ on:
...
-        with:
-          command: clippy
-          # Set linting rules for clippy
-          args: --all-targets --all-features
+        - uses: actions/checkout@v4
+        - uses: dtolnay/rust-toolchain@stable
+        - uses: Swatinem/rust-cache@v2
+        - run: cargo fmt --all -- --check
+        - run: cargo check --workspace --locked
+        - run: cargo clippy --all-targets --all-features
```
coffee-cup, line 21: Bit nit: could you please add `name`s to all these steps (same with the other steps)

.DS_Store
coffee-cup, line 1: This file should be ignored

.github/workflows/release.yml
coffee-cup: Yup agreed it isn't ideal to run locally. So the flow will be
- Manually push a new Git tag
- Release action is triggered which creates a new GH release based on tag version and builds binaries for all platforms
- Publish action is triggered on release which publishes a new version to crates.io
?

Once this flow is finalized could you please document it in the [contributing.md](https://github.com/railwayapp/cli/blob/master/CONTRIBUTING.md) file.

coffee-cup: One benefit of `cargo release` is that it would automatically bump the version in `Cargo.toml`. Will that be a manual process now?

Milo123459: Nayural process now, and the tag is automaticalmy crated.

coffee-cup: I might be missing something, but what automatically creates the tag?

Milo123459: [- name: Create GitHub release](https://github.com/railwayapp/cli/blob/f0acca3c0c8db89e49baf8820037f3fc31c4fb15/.github/workflows/release.yml#L40-L47)

.github/workflows/ci.yml
```diff
@@ -27,7 +33,11 @@ jobs:
...
     steps:
-        - uses: actions/checkout@v4
-        - uses: dtolnay/rust-toolchain@stable
-        - uses: Swatinem/rust-cache@v2
-        - run: cargo test
+        - name: checkout
+          uses: actions/checkout@v4
+        - name: install rust
+          uses: dtolnay/rust-toolchain@stable
+        - name: cahce dependencies
```
zuchka, line 40: small typo: cahce --> cache

Milo123459:
im currently testing this right now on my fork. creating a regular commit named `release v3.9.0` doesn't work, and you need to manually make a tag for some reason.

.github/workflows/release.yml
coffee-cup: That workflow runs when there is a new tag, https://github.com/railwayapp/cli/blob/f0acca3c0c8db89e49baf8820037f3fc31c4fb15/.github/workflows/release.yml#L5. What creates the new tag?

Milo123459: `git tag vX.X.X`
`git push origin vX.X.X`

Milo123459: yeah.. this is starting to get a bit annoying. I'm going to check how I do it in my CI.

aleksrutins:
Would this fix #494

PR #506, opened 2024-06-05
`railway up --ci` now waits for healthchecks

### LINKED ISSUE #503: CI Flag Should Wait for Healthchecks
The --ci flag should wait for healthchecks before exiting (and should fail if the healtchecks fail)

Milo123459:
`railway up --ci` now waits for healthchecks.

Fixes #503.

This also removes the annoying polling method of finding out if the deployment succeeded, switching to a subscription.

### DIFF
--- a/q/Dockerfile
+++ b/q/Dockerfile
@@ -1,0 +1,1 @@
+FROM dbgate/dbgate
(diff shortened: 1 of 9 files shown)


JakeCooper (approved):
Huge!

PR #508, opened 2024-06-06
Fix service command showing services that aren't in current environment

### LINKED ISSUE #507: `railway service` shows all services in project, not just the ones in the current environment

Milo123459:
Fixes #507.

This makes sure the service command only displays services that are in the current environment.


zuchka (approved):
lgtm!

PR #511, opened 2024-06-11
Fix bad formatting

Milo123459:
This pull request fixes a formatting issue in the `up.rs` file. The `Build failed` message was not being displayed correctly. This PR updates the message to display the correct error information and adds a fallback message if the error is empty.


ndneighbor (approved):
👍

PR #514, opened 2024-06-24
Fix CLI compilation

paulocsanz:
Updated rust version in nix shell.

Updated packages and fixed compilation error caused by breaking changes from hyper 1.0.0-rc, since cargo install doesn't respect cargo lock.

Added the SystemConfiguration framework on mac, since the linker was failing because of it in my machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /target
 node_modules
 bin/railway
+.envrc
+.direnv/*
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,39 +3,34 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.22.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler"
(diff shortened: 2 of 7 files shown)


paulocsanz:
It's not clear to me which release level this is since it's unblocking compilation of previous changes that maybe aren't released (?)

JakeCooper:
Shouldn’t need a release label. That’s only for “external releases”

But you can put patch too

src/util/hyper.rs
```diff
@@ -0,0 +1,154 @@
+///! Hyper compat layer after traits got split from tokio's
```
JakeCooper, line 1: Seems odd that we need this especially with the unsafe

paulocsanz: Will take a look, but that's how hyper does it for the examples now. It seems to be a recent, badly documented change. I will investigate more, hadn't seen the unsafe, was just trying to unblock myself from exploring the cli, my bad

paulocsanz: Ah, although they inline that logic in their examples and were recommending that in issues, apparently now it's part of hyper_util, so that should do it! https://docs.rs/hyper-util/latest/hyper_util/rt/tokio/struct.TokioIo.html

paulocsanz: Done, removed the inlining of that logic, using the one from hyper_util!

JakeCooper (approved):

PR #524, opened 2024-08-04
Don't prompt for environment in `railway link` if only one is present

Milo123459:
If the project being linked only has one environment, we won't prompt for it and default to the only environment there is.


paulocsanz:
Oh sorry, I hadn't seen it and ended up implementing it already. Thank you for all the PRs though! We are updating our linear to automatically notify us about PRs and issues in a better way so this does not happen again.

PR #530, opened 2024-08-14
Fix Docker publish workflow

maddsua:
This completely overhauls the workflow bringing next changes:

1. Build artifacts are reused. Instead of building the cli twice, now it utilizes build artifacts from the "Release" workflow
2. The workflow is now triggered on successful completion of "Release" instead of release publish event
3. Removed "cosign" ~which was the primary issue behind previews version of this workflow failing. Personally I don't see a need in this, but if someone thinks otherwise we can work on fixing this as well~ as it requires a signing key to be provided. It does not work with "github magic" and @alexng353, who added it to the original workflow somehow missed that. So did I. What it does when no key provided is literally waits for user to open a link in a browser and validate their identity. Which ain't happening in automated runs that this repo uses.


maddsua:
Also wanted to mention that tags are trimmed to omit "v**" at the beginning, so that a tag "v4.0.0" would become "4.0.0". It just seems to be more in line with docker conventions

ndneighbor:
The CI flag should be running with a API key fwiw so it should fail if there is no auth. I can so some testing there to make sure that the behavior is as expected.

ndneighbor (approved):
I can't really see a reason on why you would want to build the Railway CLI to deploy so this is welcome but I am afraid someone out there will bring it up as missed behavior. 

Either way, this passes the smell test.

maddsua:
The CI flag? Im not sure why you are bringing it up

maddsua:
> I can't really see a reason on why you would want to build the Railway CLI (twice) to deploy so this is welcome but I am afraid someone out there will bring it up as missed behavior.

Here's the thing, I DON'T want to rebuild cli for docker specifically so I reuse binary artifact from Release workflow

ndneighbor:
> > I can't really see a reason on why you would want to build the Railway CLI (twice) to deploy so this is welcome but I am afraid someone out there will bring it up as missed behavior.
> 
> Here's the thing, I DON'T want to rebuild cli for docker specifically so I reuse binary artifact from Release workflow

Oh I know! I just fear someone randomly being: "hey I was using that!" - you are good :)

PR #531, opened 2024-08-14
Fix exec permission in docker image

maddsua:
Fixes this whoopsie daisy:

fucky-wucky

TLDR: github artifacts reset file permissions and as someone who's used to gitlab I didn't even think of checking them. Whoops! Added a chmod command and now.. now it's done.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
+# This dockerfile is meant to be used in publish-docker.yml workflow which reuses artifacts created by release.yml
+# This way we can avoid rebuilding the CLI specifically for docker
+
 FROM alpine:3.20
 
-COPY ./railway /usr/bin/railway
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,10 @@
...
+# This way we can avoid rebuilding the CLI specifically for docker
+
 FROM alpine:3.20
 
+# binary should be retrieved by download-artifact action by this point
 COPY ./railway /usr/bin/railway
-RUN test -f /usr/bin/railway
+
+# ensure it has exec permission
+RUN chmod +x /usr/bin/railway
```
brody192, line 10: To save a line, use the chmod flag in the COPY instruction.

maddsua: done

coffee-cup (approved):

PR #536, opened 2024-08-15
Fix CI mode immediate exit

maddsua:
Fixes premature "up" command exit in non-interactive environemnts

## Cause:
An `if` statement at:
https://github.com/railwayapp/cli/blob/f2d2b07238d2814fa2f85807bbfb6ee61bb2d64a/src/commands/up.rs#L323
doesn't take into account that CI environments would be detected as "not terminal" which in turn short circuits the CI mode.

## Changes

1. Check for CI mode as well before exiting if "not in terminal".
2. Detect `CI=true` env variable that is set by every major CI provider/runner.
3. Log "CI mode enabled" message to clearly indicate current command behavior.
4. CLI in CI mode will exit with error if it can't stream build logs
5. Updated readme and help strings to reflect changes to CI mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ yay -S railwayapp-cli
 
 ### Docker
 
-Before using CLI in non-interactive environment make sure that you have created an access token (account or project scoped) and made it available via the `RAILWAY_TOKEN` environment variable.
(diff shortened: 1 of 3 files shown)


brody192:
Closing in favor of #537

PR #537, opened 2024-08-20
Fix up command CI mode and status subscription

maddsua:
This updates graphql schema and switches for a different deploy status check subscription.

This also includes changes from #536 

~**Todo: railway team needs to fix project token auth first**.
Context: `deployment` subscription does not work with project access tokens~


maddsua:
Would be noice to get #536 merged first as that is a separate bug and it would make more sense to do it this way

README.md
```diff
@@ -64,7 +64,7 @@ yay -S railwayapp-cli
 
 ### Docker
 
-Before using CLI in non-interactive environment make sure that you have created an access token (account or project scoped) and made it available via the `RAILWAY_TOKEN` environment variable.
+Before using the CLI in a non-interactive environment, ensure you have created an access token (account or project-scoped) and set it as the `RAILWAY_TOKEN` environment variable. CI environments are automatically detected by the presence of `CI=true` variable. In these environments, only build logs will be streamed, and the CLI will exit with an appropriate code indicating success or failure.
```
brody192, line 67: Account and Project tokens are named differently -

`RAILWAY_TOKEN` & `RAILWAY_API_TOKEN`

https://github.com/railwayapp/cli/blob/master/src/config.rs#L111
https://github.com/railwayapp/cli/blob/master/src/config.rs#L115

maddsua: Updated readme to only mention project-tokens as we can't really use account-scoped tokens rn (missing project id variable support)

brody192: Okay, that sounds good, we can tackle that at a later date.

brody192 (approved):

coffee-cup (approved):
thanks for the PR!

PR #542, opened 2024-09-12
Handle null values in service variables

FarazPatankar:
This PR updates several CLI commands to not fail if a service variable value is `null`. This can happen if the variable is a sealed variable.

The core change is that `get_service_variables` filters out variables with a `null` value and the remaining changes are replacing usages of the query that didn't re-use the function to re-use the function.

Note: Leaving as a draft until reviewed by someone else from the team that actually knows Rust. ^ _ ^


coffee-cup (approved):
🎉

paulocsanz (approved):

PR #546, opened 2024-09-21
show next steps when linking a project

coffee-cup:
Currently, after linking a project we don't give much feedback. This PR updates that to show the project that was linked and a few commands that you can run.

image

### DIFF
--- a/src/commands/link.rs
+++ b/src/commands/link.rs
@@ -139,6 +139,19 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
 
     configs.write()?;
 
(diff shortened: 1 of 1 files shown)


brody192:
Curious on the reason for this PR? Has there been a lot of confusion? 

But for real, as someone who knows what to do, I don't want to be told, and for someone who might not know what to do, they might not even need those commands.

Similarly to how we removed a log telling the user to run `railway up` after `railway run` exited, I don't think the cli should be telling the users how to use it, that should be left to the docs - https://docs.railway.app/quick-start#deploying-your-project---with-the-cli

PR #547, opened 2024-09-23
Allow setting variables via `--set`

Milo123459:
This allows you to set variables to a service. For example:

```
railway variables --set "HELLO=WORLD" --set "MULTIPLE=AT_ONCE"
```

### DIFF
--- a/src/commands/variables.rs
+++ b/src/commands/variables.rs
@@ -67,53 +67,15 @@ pub async fn command(args: Args, json: bool) -> Result<()> {
     };
 
     if !args.set.is_empty() {
-        let variables: BTreeMap<String, String> = args
-            .set
-            .iter()
-            .filter_map(|v| {
-                let mut split = v.split('=');
-                let key = split.next()?.trim().to_owned();
-                let value = split.collect::<Vec<&str>>().join("=").trim().to_owned();
-                if value.is_empty() {
-                    None
-                } else {
-                    Some((key, value))
-                }
-            })
-            .collect();
-
-        let fmt_variables = variables
-            .keys()
-            .map(|k| k.bold().to_string())
-            .collect::<Vec<String>>()
-            .join(", ");
-
-        let spinner = indicatif::ProgressBar::new_spinner()
-            .with_style(
-                indicatif::ProgressStyle::default_spinner()
-                    .tick_chars(TICK_STRING)
-                    .template("{spinner:.green} {msg}")
-                    .expect("Failed to set spinner template"),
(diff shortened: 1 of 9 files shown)


src/commands/variables.rs
```diff
@@ -1,21 +1,40 @@
...
+    environment: Option<String>,
+
     /// Show variables in KV format
     #[clap(short, long)]
     kv: bool,
+
+    /// The "{key}={value}" environment variable pair to set the service variables.
+    /// Example:
+    ///
+    /// ```bash
```
coffee-cup, line 33: The quotes shouldn't be here as it doesn't render properly with `--help`

image

src/commands/variables.rs
```diff
@@ -24,53 +43,88 @@ pub async fn command(args: Args, json: bool) -> Result<()> {
...
+            configs.get_backboard(),
+            vars,
         )
-    } else {
-        return Err(RailwayError::NoServiceLinked.into());
-    };
+        .await?;
+
+        spinner.finish_with_message(format!("Set {fmt_variables}"));
+        return Ok(());
```
coffee-cup, line 119: Suggestion: Can this be extracted out to its own function. We could then also have another function to parse and format the variables from the args that is testable

Milo123459: I don't think that would be useful. This logic only applies to the --set part.

ndneighbor: Less so about the usefulness in the command writ large but more so, it's easier from a maintenance perspective to have the function separate so if you or the team adds this elsewhere, we don't have to chase down a bespoke implementation.

Milo123459: Fixed

src/commands/variables.rs
```diff
@@ -91,8 +105,64 @@ pub async fn command(args: Args, json: bool) -> Result<()> {
...
+    spinner.enable_steady_tick(Duration::from_millis(100));
+    let vars = mutations::variable_collection_upsert::Variables {
+        project_id: project,
+        environment_id,
+        service_id,
+        variables,
+    };
+    post_graphql::<mutations::VariableCollectionUpsert, _>(client, configs.get_backboard(), vars)
+        .await?;
+    spinner.finish_with_message(format!("Set {fmt_variables}"));
```
coffee-cup, line 166: Very big nit
```suggestion
    spinner.finish_with_message(format!("Set variables {fmt_variables}"));
```

coffee-cup (approved):

PR #549, opened 2024-09-23
Add `railway env` as an alias for `railway environment`

Milo123459:
Alias' can now be created like this:

```rs
commands_enum!(
 actual_name(aliases, goes, here)
)
```

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,27 +1,34 @@
 #[macro_export]
 macro_rules! commands_enum {
-    ($($module:ident),*) => (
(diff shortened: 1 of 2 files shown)


coffee-cup (commented):
Curious on the reason for this PR? Has there been a lot of confusion?

brody192:
Its easier to type out `env` than `environment` on terminals that don't have completion, just a quality of life improvement in my opinion.

Context, I've asked milo for this.

coffee-cup (approved):

PR #563, opened 2024-10-04
Template deploy v2

coffee-cup:
- Use a `schema.json` file instead
- Add instructions for re-generating the schema in the contributor docs
- Add a `build.rs` file so that we rebuild on schema or graphql changes
- Update the template `deploy` and `add` commands to use the `templateDeployV2` mutation


Milo123459:
why move to a json schema?

coffee-cup:
> why move to a json schema?

Because that is what the graphql-client introspect schema returns

PR #570, opened 2024-10-24
Update `railway add`

Milo123459:
This allows users to create services and databases in the linked project and environment, with similar options to the `+ Create` menu found within the project canvas.

Changes and improvements are outlined below.

image

image

image

### DIFF
--- a/src/commands/add.rs
+++ b/src/commands/add.rs
@@ -17,7 +17,7 @@ use crate::{
 
 use super::*;
 
-/// Provision a database into your project
+/// Add a service to your project
 #[derive(Parser)]
 pub struct Args {
     /// The name of the database to add
--- a/src/commands/add.rs
+++ b/src/commands/add.rs
@@ -1,11 +1,14 @@
 use anyhow::bail;
 use is_terminal::IsTerminal;
-use std::collections::HashMap;
-use strum::IntoEnumIterator;
+use std::collections::{BTreeMap, HashMap};
+use strum::{Display, EnumIs, EnumIter, IntoEnumIterator};
 
 use crate::{
     controllers::{database::DatabaseType, project::ensure_project_and_environment_exist},
-    util::prompt::prompt_multi_options,
+    util::prompt::{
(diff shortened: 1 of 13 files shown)


coffee-cup (commented):
Could we reuse the `railway add` command to cut down on the number of commands a user has to remember.

We could add options for empty service, docker image, and github repo. Right now `railway create` just makes an empty service and it would be nice if you could do something like `railway add --name my-service --repo {pasted github link}`. Or if you just do `railway add` you are prompted all the way through the process.

brody192:
I agree having `add` as a single way of adding any kind of service(s) -- eventually bringing feature parity with the `+ Create` modal in the project canvas.

For now, I think the format of `railway add --service hello-world` would fit best with how `railway add` currently works.

`railway add --service` -> Creates an empty service with a random name.
`railway add --service hello-world` -> Creates an empty service with the specified name.

And you can provide either a `--repo` or an `--image` flag to set the source and deploy, omitting them creates an empty service.

Additionally, you would also be able to pass `--variables` to the `--service` flag.

For the interactive flow when running just `railway add` it should prompt similarly to how the `+ Create` button does.

`GitHub Repo`
`Database`
`Docker Image`

We can worry about templates later.

src/commands/add.rs
```diff
@@ -1,53 +1,337 @@
...
+/// Add a service to your project
 #[derive(Parser)]
 pub struct Args {
     /// The name of the database to add
     #[arg(short, long, value_enum)]
     database: Vec<DatabaseType>,
+
+    /// The name of the service to create (leave blank for randomly generated)
+    #[clap(short, long)]
+    service: Option<Option<String>>,
```
paulocsanz, line 29: Whats up with the nested options?

paulocsanz (approved):

src/commands/add.rs
Milo123459: To allow for --service to be specified without a value, allowing for randomly generated service names.

Issue #512, opened 2024-06-19
Add CLI feature for adding a custom domain


jph00:
It would be helpful to have something in the CLI to allow us to add a custom domain. AFAICT current `railway domain` only adds an internal one. I'd like to be able to automate the process of attaching a custom domain.

JakeCooper:
I'm happy to throw a community bounty up on this one for $100. Payable via GH sponsors/BMAC/etc

Mihir867:
Hey, Can you assign this to me ?

JakeCooper:
I think this can go to whoever has a working PR setup for it. Happy to pay out whoever

PR #571, opened 2024-11-04
[WIP] Feat: Support for adding Custom Domains with `railway domain [domain] [-p | --port]`

alexng353:
Usage: `railway domain [domain] [-p | --port]`

~~Also can run interactively. Currently am unsure if it will work headless, but willing to test if that's necessary for this feature.~~

The command runs entirely headlessly. If you are creating a custom domain in headless mode, I would recommend using the `--json` flag to get a clean JSON output of the DNS records.

Note: I updated the GQL schema using the method provided in [CONTRIBUTING.md](https://github.com/railwayapp/cli/blob/master/CONTRIBUTING.md), which can be reverted if the admin deems it necessary (for whatever reason).

Docs PR: https://github.com/railwayapp/docs/pull/613

Fixes: #512 

image


src/commands/domain.rs
```diff
@@ -148,3 +139,48 @@ fn print_existing_domains(domains: &DomainsDomains) -> Result<()> {
 
     Ok(())
 }
+
+// Returns a reference to save on Heap allocations
+pub fn get_service<'a>(
+    linked_project: &'a LinkedProject,
+    project: &'a queries::project::ProjectProject,
+) -> Result<&'a queries::project::ProjectProjectServicesEdgesNode, anyhow::Error> {
```
Milo123459, line 147: nit: use anyhow::Result type

alexng353: fixed

src/commands/domain.rs
```diff
@@ -13,12 +14,33 @@ use crate::{
...
 
-/// Generates a domain for a service if there is not a railway provided domain
-// Checks if the user is linked to a service, if not, it will generate a domain for the default service
+/// Add a custom domain or generate a railway provided domain for a service.
+///
+/// There is a maximum of 1 railway provided domain per service.
 #[derive(Parser)]
-pub struct Args {}
+pub struct Args {
+    /// The service to generate a domain for
```
coffee-cup, line 22:
```suggestion
    /// The port to connect to the domain
```

src/commands/domain.rs
```diff
@@ -13,12 +14,33 @@ use crate::{
...
+    #[clap(short, long)]
+    port: Option<u16>,
+
+    /// The name of the service to use
```
coffee-cup, line 26:
```suggestion
    /// The name of the service to generate the domain for
```

src/commands/domain.rs
```diff
@@ -13,12 +14,33 @@ use crate::{
...
+    /// Optionally, specify a custom domain to use. If not specified, a domain will be generated.
+    ///
+    /// Specifying a custom domain will also return the required DNS records
+    /// to add to your DNS settings
+    domain: Option<String>,
+}
+
+pub async fn command(args: Args, json: bool) -> Result<()> {
+    if let Some(domain) = args.domain {
+        create_custom_domain(domain, args.port, args.service, json).await?;
```
coffee-cup, line 39: Can you also move the service domain creation to a `create_service_domain` function so it is clear what type of domain is being created.

src/commands/domain.rs
```diff
@@ -148,3 +135,196 @@ fn print_existing_domains(domains: &DomainsDomains) -> Result<()> {
...
+        // and in that case, the post_graphql call should have already errored.
+        bail!("No DNS records found. Please check the Railway dashboard for more information.");
+    }
+
+    println!(
+        "To finish setting up your custom domain, add the following DNS records to {}:\n",
+        &response.custom_domain_create.status.dns_records[0].zone
+    );
+
+    print_dns(response.custom_domain_create.status.dns_records);
```
coffee-cup, line 274: Will there ever be more than 1 DNS record? It might make it clearer to just `print_dns` with a single record at a time in a loop for the very edge case where there are more than 1 record.

src/commands/domain.rs
```diff
@@ -148,3 +135,196 @@ fn print_existing_domains(domains: &DomainsDomains) -> Result<()> {
...
+
+    Ok(())
+}
+
+fn print_dns(
+    domains: Vec<
+        mutations::custom_domain_create::CustomDomainCreateCustomDomainCreateStatusDnsRecords,
+    >,
+) {
+    // I benchmarked this iter().fold() and it's faster than using 3x iter().map()
```
coffee-cup, line 287: I think this is fine, but just for future reference, readability is much more important here than performance. Especially considering the `domains` length will be size 1 99% of the time.

coffee-cup (approved):
Looking good! Just a few comments

src/commands/domain.rs
alexng353: Yes, if you use a `*` wildcard it will have 2 records.
image

src/commands/domain.rs
alexng353: Yeah, that makes sense. For `n=1`, the performance is entirely irrelevant.

src/commands/domain.rs
alexng353: Ok

src/commands/domain.rs
coffee-cup: Even for `n=2`. The difference will be so small it is negligible. Unless thousands of domains are being processed, readability is what we should optimize for.

coffee-cup: I think it is fine as is, just mainly a note for the future

coffee-cup (approved):
🎉

PR #573, opened 2024-11-15
Make service optional when linking

Milo123459:
This allows you to press escape to skip providing a service when running railway link interactively.

Why: Sometimes people may want to link to a project for a monorepo without specifically selecting one service.


samoriental:
This is the change of the century. Thank you.

coffee-cup (approved):

PR #577, opened 2024-12-02
feat: add check updates command

alexng353:
This PR adds automatic (daily) checks for new updates for the CLI, only in interactive mode. Alternatively, a user may manually check for updates using the `railway check-updates` command. 

Automatic daily checks run in interactive mode, in the background, for any command called. If a new version is detected, the CLI will push a message to the user (similar to NPM) that requests they update their CLI. 

Manual checks can be run as follows:
```bash
railway check-updates
```

This is what the check-updates command looks like when the version has changed:
image

This is what the check-updates command looks like when the version has not changed:
image

This is what the automatic check looks like when the check detects that the version has changed:
image

The data is grabbed (rather slowly) via the GitHub API (endpoint https://api.github.com/repos/railwayapp/cli/releases/latest), which takes around 700ms to process. This is why we run the check in the background, and even then, we only run it daily.

First run:
image
Second run:
image

### DIFF
--- a/src/commands/check_updates.rs
+++ b/src/commands/check_updates.rs
@@ -1,4 +1,4 @@
-use crate::check_update;
+use crate::util::check_update::check_update;
 
 use super::*;
 use serde_json::json;
@@ -23,7 +23,7 @@ pub async fn command(_args: Args, json: bool) -> Result<()> {
         return Ok(());
     }
 
-    let is_latest = check_update!(configs, true);
+    let is_latest = check_update(&mut configs, true).await?;
     if is_latest {
         println!(
             "You are on the latest version of the CLI, v{}",
--- a/src/commands/check_updates.rs
+++ b/src/commands/check_updates.rs
@@ -13,22 +13,22 @@ pub async fn command(_args: Args, json: bool) -> Result<()> {
     if json {
         let result = configs.check_update(true).await;
 
-        if let Ok(Some(latest_version)) = result {
-            let json = json!({
-                "latest_version": latest_version,
-                "current_version": env!("CARGO_PKG_VERSION"),
-            });
-            println!("{}", serde_json::to_string_pretty(&json)?);
-        } else {
-            let json = json!({
-                "latest_version": None::<String>,
-                "current_version": env!("CARGO_PKG_VERSION"),
-            });
-            println!("{}", serde_json::to_string_pretty(&json)?);
-        }
+        let json = json!({
+            "latest_version": result.ok().flatten().as_ref(),
+            "current_version": env!("CARGO_PKG_VERSION"),
+        });
+
+        println!("{}", serde_json::to_string_pretty(&json)?);
+
         return Ok(());
     }
 
-    check_update!(configs, true);
+    let is_latest = check_update!(configs, true);
+    if is_latest {
+        println!(
+            "You are on the latest version of the CLI, v{}",
+            env!("CARGO_PKG_VERSION")
+        );
+    }
     Ok(())
 }
--- a/src/commands/check_updates.rs
+++ b/src/commands/check_updates.rs
@@ -1,0 +1,34 @@
+use crate::check_update;
+
+use super::*;
+use serde_json::json;
+
+/// Test the update check
+#[derive(Parser)]
+pub struct Args {}
+
+pub async fn command(_args: Args, json: bool) -> Result<()> {
+    let mut configs = Configs::new()?;
+
+    if json {
+        let result = configs.check_update(true).await;
+
+        if let Ok(Some(latest_version)) = result {
+            let json = json!({
+                "latest_version": latest_version,
+                "current_version": env!("CARGO_PKG_VERSION"),
+            });
+            println!("{}", serde_json::to_string_pretty(&json)?);
+        } else {
+            let json = json!({
+                "latest_version": None::<String>,
+                "current_version": env!("CARGO_PKG_VERSION"),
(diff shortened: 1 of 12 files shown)


src/commands/check_updates.rs
```diff
@@ -0,0 +1,34 @@
+use crate::check_update;
+
+use super::*;
+use serde_json::json;
+
+/// Test the update check
+#[derive(Parser)]
+pub struct Args {}
+
+pub async fn command(_args: Args, json: bool) -> Result<()> {
```
coffee-cup, line 10: We should output if we are on the latest version. Right now it doesn't output anything and is a bit confusing

image

coffee-cup (commented):
Looking good and just a few comments.

Can you also please add descriptions to any PRs so it makes it easier to review. This should include things like
- Added commands (screenshots of the output is very useful)
- Changes to behaviour.

src/macros.rs
```diff
@@ -42,3 +42,22 @@ macro_rules! interact_or {
         }
     };
 }
+
+#[macro_export]
+macro_rules! check_update {
```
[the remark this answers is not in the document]
alexng353: It's a macro because I can overload the call signature of it to default the second parameter to `false`, but I can make it into a function if you want.

src/config.rs
```diff
@@ -313,4 +326,43 @@ impl Configs {
...
+        let client = reqwest::Client::new();
+        let response = client
+            .get(GITHUB_API_RELEASE_URL)
+            .header("User-Agent", "railwayapp")
+            .send()
+            .await?;
+
+        self.root_config.last_update_check = Some(Utc::now());
+        self.write()
+            .context("Failed to write config in should_update")?;
```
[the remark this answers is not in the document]
alexng353: `"Failed to time since last update check"`

src/macros.rs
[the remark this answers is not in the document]
coffee-cup: I think a function is preferred here for simplicity. If you want to avoid having to provided `false` as the second parameter, we can have a `check_update_command` function that passed false

src/commands/check_updates.rs
```diff
@@ -0,0 +1,34 @@
+use crate::check_update;
+
+use super::*;
+use serde_json::json;
+
+/// Test the update check
```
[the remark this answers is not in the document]
coffee-cup: Ah sorry. Means to delete this line (or change it). "Test the update check" is not very clear

src/macros.rs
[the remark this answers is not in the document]
alexng353: That sounds good to me. I just finished finals, so I'll finalize this now.

coffee-cup (commented):
Would love to get this merged soon! Just a small merge conflict and a few linting rules but then I think it would be good to go

alexng353:
> Would love to get this merged soon! Just a small merge conflict and a few linting rules but then I think it would be good to go

Merge conflict rectified via rebase.
Lint fixed with `cargo fmt`.

src/main.rs
```diff
@@ -21,6 +22,7 @@ mod macros;
 #[derive(Parser)]
 #[clap(author, version, about, long_about = None)]
 #[clap(propagate_version = true)]
+// #[clap(author, about, long_about = None)]
```
Milo123459, line 25: could this be removed?

src/main.rs
```diff
@@ -58,11 +60,24 @@ commands_enum!(
...
     volume,
-    redeploy
+    redeploy,
+    check_updates
 );
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    // intercept the args
+    {
```
Milo123459, line 70: is there not a better way to do this w clap? seems a bit hacky

Milo123459 (approved):
looks good!

src/main.rs
alexng353: I researched (googled and ChatGPT'd) for 2 or 3 hours before writing this, and the answer seems to be unequivocally no. It's not like doing this costs anything at all, the number of instructions generated and the number of clock cycles used to check 5 vector positions are both negligible.

Milo123459 (approved):
lgtm. well done!

src/main.rs
```diff
@@ -58,27 +62,117 @@ commands_enum!(
...
+    //
+    // This code tells us what stream to print the error to:
+    // https://github.com/clap-rs/clap/blob/cb2352f84a7663f32a89e70f01ad24446d5fa1e2/clap_builder/src/error/mod.rs#L210-L215
+    //
+    // pub(crate) fn stream(&self) -> Stream {
+    //     match self.kind() {
+    //         ErrorKind::DisplayHelp | ErrorKind::DisplayVersion => Stream::Stdout,
+    //         _ => Stream::Stderr,
+    //     }
+    // }
```
coffee-cup, line 146: Can this be removed? What is it for?

coffee-cup (approved):
Looks good. Just wondering about that comment in main, but this can be merged after

src/main.rs
alexng353: I can remove it, but it's helpful for future devs if they want to look into the reasoning behind the main file's error handling

coffee-cup: I think its okay to leave a comment with a link, but this big block is unnecessary IMO.

alexng353: Will remove 👍

coffee-cup (approved):

PR #584, opened 2024-12-31
Fix hanging of --ci option and watch patterns

### LINKED ISSUE #582: `railway up --ci` does not exit if no files changed
When running `railway up` in a project with `watchPatterns` defined, if no files are changed the cli hangs on the `No changed files matched patterns: <patterns>` message instead of exiting with code 0

```
❯ railway --version
railwayapp 3.20.0
```

Milo123459:
Fixes #582

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -335,8 +335,13 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
     //	Always stream build logs
     let build_deployment_id = deployment_id.clone();
     let mut tasks = vec![tokio::task::spawn(async move {
-        if let Err(e) =
-            stream_build_logs(build_deployment_id, |log| println!("{}", log.message)).await
+        if let Err(e) = stream_build_logs(build_deployment_id, |log| {
+            println!("{}", log.message);
+            if args.ci && log.message.starts_with("No changed files matched patterns") {
+                std::process::exit(0);
+            }
+        })
+        .await
         {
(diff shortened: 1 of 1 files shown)


src/commands/up.rs
```diff
@@ -335,8 +335,13 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
     //	Always stream build logs
     let build_deployment_id = deployment_id.clone();
     let mut tasks = vec![tokio::task::spawn(async move {
-        if let Err(e) =
-            stream_build_logs(build_deployment_id, |log| println!("{}", log.message)).await
+        if let Err(e) = stream_build_logs(build_deployment_id, |log| {
+            println!("{}", log.message);
+            if args.ci && log.message.starts_with("No changed files matched patterns") {
```
maddsua, line 340: Wait is checking for a substring the best way to detect the absence of changes?

Milo123459: unfortunately, yes

brody192: Didn't you wanna wait for the skipped event instead haha

Milo123459: no because then it wouldn't log anything about what patterns were skipped, that gets sent as a string on the build logs subscription, we get the SKIPPED event always before that

brody192: Ah makes sense, sounds good to me!

coffee-cup (approved):

PR #590, opened 2025-02-03
Add alias for `railway env new` as `railway env create`

Milo123459:


src/commands/environment.rs
```diff
@@ -29,6 +29,7 @@ pub struct Args {
 #[derive(Subcommand)]
 pub enum EnvironmentCommand {
     /// Create a new environment
+    #[clap(visible_alias = "create")]
```
paulocsanz, line 32: Are we sure we want to do this? We can't remove the alias in the future, and the CLI will get bloated with redundant options that we are stuck with forever.

Can we give a nice error message on create like rust does for typos "did you mean railway env new?"

Milo123459: I'm not sure about the second bit, but I think aliasing create is fine to be honest.

PR #596, opened 2025-02-18
Adds a flag for correct uploading of project subdirs

### LINKED ISSUE #595: `railway up <path>` doesn't zip the project properly
I was exploring how to automate [wasp-lang/wasp](https://github.com/wasp-lang/wasp) deployments to Railway with Wasp's CLI (we already have it for [Fly](https://wasp-lang.dev/docs/deployment/deployment-methods/cli#flyio) and we want to support Railway next). I believe I have stumbled upon a bug in the Railway CLI in the `railway up` command.

### How to reproduce it

**Some context:** Wasp apps are full-stack apps that you compile into a Node.js server app and a React client app in different folders. You can think of deploying Wasp apps as deploying three Railway services: client, server and the database. So when we want to deploy the e.g. server app, we deploy it from the server folder.

What I did to deploy the server app:
1. Init the project with `railway init --name "my-app"` in the project root dir
2. Build the Wasp app
3. Go into the server app folder `cd .wasp/build`
4. Create a new server app service with `railway add --service "my-app-server" --variables "PORT=8080" ...`
5. Try to deploy the server app service with `railway up $(pwd) --service "my-app-server" --no-gitignore --detach`

The deployment fails with some Nixpacks related error which is strange because we have a `Dockerfile` in the `.wasp/build` folder. I was curious and I wanted to find out what's going on, so I've built the CLI from source and added some debug logs.

### Building from source and debugging

When I added logs in the `src/commands/up.rs` to see which kind of paths are being included in the zip I got output similar to this:
```
.wasp/build/Dockerfile
.wasp/build/server.js
...
```

I did not expect the zip to include the folder structure like that, I thought my dir will be zipped as if it's the root dir. 
I expected the zip to look like this:
```
./Dockerfile
./server.js
...
```

I guess that's the reason why `Dockerfile` was not picked up by Railway - it's not in the root of the zip. That didn't seem right, so I explored some more.

### Possible bug

Inspecting the [src/commands/up.rs](https://github.com/railwayapp/cli/blob/master/src/commands/up.rs#L132) a bit more, I found out that there were two paths being used in the `command` function: 

```rs
// First path
let prefix: PathBuf = configs.get_closest_linked_project_directory()?.into();

// Second path
let path = match args.path {
    Some(path) => path,
    None => prefix.clone(),
};
```
- `prefix` is the dir of the project
- `path` is either provided via args (as I do in my example above) or the same as `prefix`.

I found that the `up` command walks the `path` dir tree and finds all the files to add to the zip **but** uses `prefix` to strip the first part of the file path:
```rs
let mut builder = WalkBuilder::new(path);
...
let stripped = PathBuf::from(".").join(path.strip_prefix(&prefix)?);
archive.append_path_with_name(path, stripped)?;
```

For example:
- if the project dir is `/root/dev/project` -> `prefix` is `/root/dev/project`
- if we provide the path arg as `/root/dev/project/.wasp/build` -> `path` is `/root/dev/project/.wasp/build`
- when we want to zip the `/root/dev/project/.wasp/build/Dockerfile` -> we only strip the project dir and we end up with `.wasp/build/Dockerfile`

### Possible solution

I'm a Rust newbie, but I tried playing around and came up with this change:
```rs
// If args.path is provided - use it, otherwise use the linked project dir 
let project_files_path = match args.path {
    Some(path) => path,
    None => configs.get_closest_linked_project_directory()?.into(),
};
```
The change:
1. merges the `path` and `prefix` concepts into a single variable `project_files_path` 
2. uses the `project_files_path` in both places

I've tested it out with my use case and it fixes the issue I had. 

Let me know if this all sounds okay to you and I can submit a PR (I have the commit ready to go: https://github.com/infomiho/railway-cli/commit/f7bcce1b78ae7c50aa518a468ba72ca5ac94c9c4)

infomiho:
Fixes #595 

It enables the users to provide the `--use-path-as-archive-prefix` flag if they want to upload a project subdir _while zipped with the subdir as archive root_. This maintains the old behavior for existing users that didn't provide the flag.

@brody192 let me know if I can improve the solution / code quality

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -11,7 +11,6 @@ use gzp::{deflate::Gzip, ZBuilder};
 use ignore::WalkBuilder;
 use indicatif::{ProgressBar, ProgressFinish, ProgressIterator, ProgressStyle};
 use is_terminal::IsTerminal;
-use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use synchronized_writer::SynchronizedWriter;
 use tar::Builder;
@@ -21,16 +20,13 @@ use crate::{
     controllers::{
         deployment::{stream_build_logs, stream_deploy_logs},
         environment::get_matched_environment,
-        project::{ensure_project_and_environment_exist, get_project},
+        project::get_project,
         service::get_or_prompt_service,
     },
     errors::RailwayError,
     subscription::subscribe_graphql,
     subscriptions::deployment::DeploymentStatus,
-    util::{
-        logs::format_attr_log,
-        prompt::{prompt_select, PromptService},
-    },
+    util::logs::format_attr_log,
 };
 
 use super::*;
@@ -61,6 +57,10 @@ pub struct Args {
     no_gitignore: bool,
 
     #[clap(long)]
+    /// Use the path argument as the prefix for the archive instead of the project directory.
+    path_as_root: bool,
+
+    #[clap(long)]
     /// Verbose output
     verbose: bool,
 }
@@ -84,12 +84,8 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
     let hostname = configs.get_host();
     let client = GQLClient::new_authorized(&configs)?;
     let linked_project = configs.get_linked_project().await?;
-    let prefix: PathBuf = configs.get_closest_linked_project_directory()?.into();
 
-    let path = match args.path {
-        Some(path) => path,
-        None => prefix.clone(),
-    };
+    let deploy_paths = get_deploy_paths(&args, &linked_project)?;
 
     let project = get_project(&client, &configs, linked_project.project.clone()).await?;
 
@@ -145,7 +141,7 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
 
     {
         let mut archive = Builder::new(&mut parz);
-        let mut builder = WalkBuilder::new(path);
+        let mut builder = WalkBuilder::new(deploy_paths.project_path);
         builder.add_custom_ignore_filename(".railwayignore");
         if !args.no_gitignore {
             builder.add_custom_ignore_filename(".gitignore");
@@ -177,7 +173,8 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
                 {
                     continue;
                 }
-                let stripped = PathBuf::from(".").join(path.strip_prefix(&prefix)?);
+                let stripped =
+                    PathBuf::from(".").join(path.strip_prefix(&deploy_paths.archive_prefix_path)?);
                 archive.append_path_with_name(path, stripped)?;
             }
         } else {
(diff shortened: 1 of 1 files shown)


brody192:
We have to have a much shorter flag to enable the new behavior 😆

infomiho:
I think the shortest flag name I can come up with is `--path-as-prefix` 😄 what do you say?

brody192:
How about instead of having the flag describe what it does technically, have it describe what it does for the user?

infomiho:
Hey, I've been pondering about this for the last few days, looking at the docs...

| Screenshot 2025-02-28 at 10 32 58 |
| --- |
| `up` command [API](https://docs.railway.com/reference/cli-api) | 

I guess, I'm looking the flag to _run as if the deploy `<path>` is the project root_, given that, we could go for:
- `--subdir-as-root`
- `--path-as-root` 
- `--deploy-dir-as-project-root`
- `--relative-root`
- `--use-as-root`
- `--from-here`

This should signal to the CLI to _use the deploy dir as if it was the project root_ so the archiving behavior would match and be correct.

src/commands/up.rs
```diff
@@ -11,7 +11,6 @@ use gzp::{deflate::Gzip, ZBuilder};
 use ignore::WalkBuilder;
 use indicatif::{ProgressBar, ProgressFinish, ProgressIterator, ProgressStyle};
 use is_terminal::IsTerminal;
-use reqwest::Client;
```
infomiho, line 14: Unused imports cleanup

src/commands/up.rs
```diff
@@ -21,16 +20,13 @@ use crate::{
...
         deployment::{stream_build_logs, stream_deploy_logs},
         environment::get_matched_environment,
-        project::{ensure_project_and_environment_exist, get_project},
+        project::get_project,
         service::get_or_prompt_service,
     },
     errors::RailwayError,
     subscription::subscribe_graphql,
     subscriptions::deployment::DeploymentStatus,
-    util::{
```
infomiho, line 30: Unused imports cleanup

src/commands/up.rs
```diff
@@ -363,3 +361,32 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
+    } else {
+        let project_dir: PathBuf = linked_project.project_path.clone().into();
+        let project_path = match args.path {
+            Some(ref path) => path.clone(),
+            None => project_dir.clone(),
+        };
+        Ok(DeployPaths {
+            project_path,
+            archive_prefix_path: project_dir,
+        })
```
infomiho, line 390: Matches the current logic: use `path` if provided, use linked project dir as archive prefix.

coffee-cup (approved):
looks good!

infomiho:
@coffee-cup should I updated the Railway CLI docs as well?

PR #598, opened 2025-02-22
add `ssh` command

20k-ultra:
This adds a `railway ssh` command that creates a websocket connect to a relay server that will provide SSH capabilities to deployment instances.

`<project-name>` and `<service-name>` are required and deployment instance id is not because they are the minimum data required to find a deployment instance. `--deployment-instance` is only necessary if you have multiple replicas.

```sh
$ railway ssh --help
Connect to a service via SSH

Usage: railway ssh [OPTIONS]

Options:
  -p, --project <PROJECT>
          Project to connect to (defaults to linked project)
  -s, --service <SERVICE>
          Service to connect to (defaults to linked service)
  -e, --environment <ENVIRONMENT>
          Environment to connect to (defaults to linked environment)
  -d, --deployment-instance <deployment-instance-id>
          Deployment instance ID to connect to (defaults to first active instance)
      --json
          Output in JSON format
  -h, --help
          Print help
  -V, --version
          Print version

$ railway ssh
  Connecting to service...
root@07e29b7e760a:/app#
```

### DIFF
--- a/src/commands/ssh.rs
+++ b/src/commands/ssh.rs
@@ -1,0 +1,124 @@
+use anyhow::{Context, Result};
+use is_terminal::IsTerminal;
+use tokio::io::AsyncReadExt;
+use tokio::select;
+
+use super::*;
+use crate::{consts::TICK_STRING, controllers::terminal::TerminalClient};
+
+/// Connect to a service via SSH
+#[derive(Parser)]
+pub struct Args {
+    /// Project to connect to
+    #[arg(value_name = "project-name")]
+    project: String,
+
+    /// Service to connect to
+    #[arg(value_name = "service-name")]
+    service: String,
+
+    /// Deployment instance ID to connect to
+    #[arg(long = "deployment-instance", value_name = "deployment-instance-id")]
+    deployment_instance: Option<String>,
+}
+
+pub async fn command(args: Args, _json: bool) -> Result<()> {
+    let configs = Configs::new()?;
+    let token = configs
+        .get_railway_auth_token()
+        .context("No authentication token found. Please login first with 'railway login'")?;
+
+    let spinner = indicatif::ProgressBar::new_spinner()
+        .with_style(
+            indicatif::ProgressStyle::default_spinner()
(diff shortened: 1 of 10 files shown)


src/commands/ssh.rs
```diff
@@ -0,0 +1,124 @@
...
+    let mut client = TerminalClient::new(
+        &ws_url,
+        &token,
+        &args.project,
+        &args.service,
+        args.deployment_instance.as_deref(),
+    )
+    .await?;
+
+    if !std::io::stdout().is_terminal() {
```
20k-ultra, line 52: move this before making ws connection

src/controllers/terminal.rs
```diff
@@ -0,0 +1,259 @@
...
+                Err(e) => {
+                    if attempt == SSH_MAX_RECONNECT_ATTEMPTS {
+                        bail!(
+                            "Failed to establish connection after {} attempts: {}",
+                            SSH_MAX_RECONNECT_ATTEMPTS,
+                            e
+                        );
+                    }
+                    eprintln!(
+                        "Connection attempt {} failed: {}. Retrying in {} seconds...",
```
20k-ultra, line 87: make sure these logs go away after connection

src/commands/ssh.rs
```diff
@@ -0,0 +1,273 @@
...
+        .with_message("Connecting to service...");
+
+    spinner.enable_steady_tick(std::time::Duration::from_millis(100));
+
+    let ws_url = format!("wss://{}", configs.get_relay_host_path());
+
+    let mut client = TerminalClient::new(&ws_url, &token, &params).await?;
+
+    // Wait a moment before getting terminal size and sending events to connection
+    tokio::time::sleep(std::time::Duration::from_millis(300)).await;
```
20k-ultra, line 113: I added this because for some reason the terminal size isn't correct or the backend isn't ready yet for the send_window_size event. This allows initial window size to correctly be set on connect.

ndneighbor (approved):

PR #604, opened 2025-03-13
`railway scale`

Milo123459:
This PR adds a railway scale command that allows you to update the amount of regions and replicas running on a service.

It works using dynamically generated flags (which are generated from the `Regions` query)

image

It also works fully interactively:

image

image

The colour of the replicas text:
- red: decrease in amount
- green: increase in amount
- yellow: remaining the same


Milo123459:
Deployments still don't seem to get the updated `multiRegionConfig` even after being redeployed. The dashboard is still showing it correctly.

paulocsanz:
Hmm will take a look again, the 2fa problem should be gone tho, right?

Milo123459:
@paulocsanz there was never a 2fa problem, a few commands require 2fa which means they don't work non-interactively, but im pretty sure that's fine.

src/macros.rs
```diff
@@ -1,30 +1,72 @@
...
+                                {
+                                    use $crate::commands::$module::*;
+                                    s = get_dynamic_args(s);
+                                }
+                            }
+                            s = s.name(stringify!($module));
+                            s
+                        };
+                        // Add this subcommand into the global CLI.
+                        cmd = cmd.subcommand(sub);
```
coffee-cup, line 45: Is this macro necessary? In general, I would avoid making macros this big. Its hard to follow and not very maintainable.

src/commands/scale.rs
```diff
@@ -0,0 +1,423 @@
...
+        };
+        regions.regions.sort_by(|a, b| {
+            get_replicas_amount(b.name.clone())
+                .1
+                .cmp(&get_replicas_amount(a.name.clone()).1)
+        });
+        let regions = regions
+            .regions
+            .iter()
+            .map(|f| {
```
coffee-cup, line 113: Can we filter out non-metal regions. They are available to everyone and we are in the process of migrating everyone to metal anyway.

src/commands/scale.rs
```diff
@@ -0,0 +1,423 @@
...
+    if let Some(latest) = latest_id {
+        let spinner = indicatif::ProgressBar::new_spinner()
+            .with_style(
+                indicatif::ProgressStyle::default_spinner()
+                    .tick_chars(TICK_STRING)
+                    .template("{spinner:.green} {msg}")?,
+            )
+            .with_message("Redeploying...");
+        spinner.enable_steady_tick(Duration::from_millis(100));
+        post_graphql::<mutations::DeploymentRedeploy, _>(
```
coffee-cup, line 211: Instead of redeploying here, I think you should be calling `serviceInstanceDeployV2`. Redeploying will intentionally use the config from the deploy you are redeploying. This is why the updated multi region config is not being picked up. If you use service instance deploy, it will use the current service config and deploy the latest commit/image/up deploy.

image

src/macros.rs
Milo123459: This macro was always how commands are defined

alexng353:
Are the region flags going to stay the same (including the relatively difficult-to-type endings for metal regions)? I'm assuming this isn't a concern at the moment because the interactive mode exists.

Milo123459:
all good on my side!

Issue #608, opened 2025-03-23
`--no-gitignore` option in `railway up` has no effects


LER0ever:
## Current Behavior
Currently the ignore patterns are handled by the `ignore` crate from ripgrep. But that crate by default takes care of `.gitignore` file through standard filters [1] and no matter if the user enables "--no-gitignore" or not, the files in the `.gitignore` will never be in the final deployment archive. 

See also the option `git_ignore` [2] (enabled by default) for the `WalkBuilder`, which can be used to control whether ripgrep/ignore respect the `.gitignore` file patterns. I can file a PR doing exactly that to fix it, already verified locally.

Judging by glancing through the history of Railway CLI and the `ignore` crate, the `--no-gitignore` flag likely has never worked as expected since its introduction more than one year ago. 

## Expected Behavior
If user enables `--no-gitignore`, then `.gitignore` should be actually bypassed, and everything in the CWD should be zipped and uploaded to Railway, allowing for use cases such as binary deployment, or internal packages that are pre-fetched in CI.

## How to reproduce

Dockerfile:

```docker
FROM alpine:latest

COPY some_binary /app/some_binary

ENTRYPOINT ["/app/some_binary"]
```

`.gitignore`:

```gitignore
some_binary
```

Command:

```shell
railway up --no-gitignore
```

This will cause a deployment failure at build stage, saying 

```
failed to calculate checksum of ref xxx: "/some_binary" not found
```

[1]: https://docs.rs/ignore/0.4.23/ignore/struct.WalkBuilder.html#method.standard_filters
[2]: https://docs.rs/ignore/0.4.23/ignore/struct.WalkBuilder.html#method.git_ignore

coffee-cup:
Fixed in https://github.com/railwayapp/cli/pull/609

PR #609, opened 2025-03-23
fix: make –no-gitignore actually work

LER0ever:
The ignore crate from ripgrep, by default, includes standard filters which automatically ignore files listed in .gitignore and other similar files. Simply not adding .gitignore manually through add_custom_ignore_filename has no practical effect, causing files listed in .gitignore to remain ignored.

This PR resolves this issue by explicitly disabling the git_ignore handler in the standard filters. This behavior has been tested locally to confirm that it correctly bypasses .gitignore.

See issue #608


coffee-cup (approved):
thanks!

PR #610, opened 2025-03-25
Properly show external workspaces

paulocsanz:
We are going away with the concept of a personal workspace. Right now personal workspaces in the CLI have both projects the user is a member of and projects owned by the workspace created for the user when they signed up.

And that workspace created for the user when they signup now has a team, so it appears twice in the selector.

To solve this we stop using `me.projects` and go throught he externalWorkspaces query, displaying each external project in its workspace.

### DIFF
--- a/src/commands/link.rs
+++ b/src/commands/link.rs
@@ -1,5 +1,7 @@
+use chrono::{DateTime, Utc};
 use colored::*;
 use is_terminal::IsTerminal;
+use serde::Serialize;
 use std::fmt::Display;
 
 use crate::{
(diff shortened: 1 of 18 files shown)


src/commands/link.rs
```diff
@@ -143,176 +143,116 @@ fn select_environment(
...
+            } else if let Some(w) = response.me.workspaces.iter().find(|w| {
+                w.team.as_ref().map_or(false, |t| {
+                    t.projects.edges.iter().any(|pro| {
+                        pro.node.id.to_lowercase() == project.to_lowercase()
+                            || pro.node.name.to_lowercase() == project.to_lowercase()
+                    })
+                })
+            }) {
+                fake_select("Select a workspace", &w.name);
+                Workspace::Member(w.clone())
```
coffee-cup, line 199: Can this be combined with an `extend` like in `init`? We might want to abstract fetching the workspaces so commands don't have to worry about external/non-external workspaces. I think a `workspace.rs` file in the controllers ([like we have for the project](https://github.com/railwayapp/cli/blob/master/src/controllers/project.rs)) would be useful

src/commands/link.rs
```diff
@@ -143,176 +129,84 @@ fn select_environment(
...
+                fake_select("Select a workspace", workspace.name());
+                workspace.clone()
             } else {
-                prompt_teams(me)?
+                prompt_workspaces(workspaces)?
             }
         }
         (None, Some(team_arg)) | (Some(_), Some(team_arg)) => {
-            match team_arg.to_lowercase().as_str() {
-                "personal" => {
```
paulocsanz, line 243: Wondering what to do about this. It's technically a regression removing the option to set the personal workspace like that. But it's not a thing anymore, so we wont be able to infer a personal workspace pretty soon.

coffee-cup (approved):

PR #613, opened 2025-03-30
Remove 2fa

Milo123459:
This PR removes 2fa from all commands, as it is no longer required through the API.


coffee-cup:
2FA is definitely still needed and used. Closing this out

PR #614, opened 2025-04-04
Impl automatic reconnect when idle in SSH

20k-ultra:
The websocket connection used for SSH can unexpectedly close for several reasons which are being improved but for now we can simply reconnect if the session has been idle for 5 seconds. This means if you have the session open and switching between windows and get a disconnect, you won't even notice. The terminal does not try to. reconstruct your previous state. For example, if you have half a command typed, this will be cleared. It's exactly like if you just manually typed `railway ssh` again.

The reconnect will be attempted every 1 second for 10 seconds. Most of this is transparent to the platform specific code (unix/windows) which makes using the ssh terminal easier. The terminal just emits a "re-connected" message that those platforms than use and clear the current terminal/etc.

Bash history isn't restored since the reconnect actually spawns another session. This means pressing your up arrow might not show the exact same history of commands. This was done so we can provide a quicker impl since a better reconnect version has been discussed but requires more work. 

Because of this future work I added a new event to the connection handshake. After the CLI connects and receives the `welcome`, it will either `init_shell` or `exec_command`. `init_shell` is used for an interactive session where you type and get data back but the process being spawned might not be ready before you type so now the client waits for a `ready` message. `exec_command` does not get a `ready`, instead it gets a `stand_by` to indicate "something is going on and data will be sent when complete". In the future we could add logic so that before we close the session we check if we are in standby mode waiting for a command to complete and tell the user that something is happening and to not quit yet.

### DIFF
--- a/src/commands/ssh/mod.rs
+++ b/src/commands/ssh/mod.rs
@@ -6,9 +6,12 @@ use crate::config::Configs;
 
 pub const SSH_CONNECTION_TIMEOUT_SECS: u64 = 30;
 pub const SSH_MESSAGE_TIMEOUT_SECS: u64 = 10;
-pub const SSH_MAX_RECONNECT_ATTEMPTS: usize = 3;
-pub const SSH_RECONNECT_DELAY_SECS: u64 = 5;
+pub const SSH_MAX_CONNECT_ATTEMPTS: usize = 3;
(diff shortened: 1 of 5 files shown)


20k-ultra:
This PR is implement reconnect and `ready`/`stand_by` logic. I know I want the latter so closing this to add in another PR while I consider if I want to add the reconnect logic.

PR #621, opened 2025-04-24
`railway run --remote`

Milo123459:
This command adds the `--remote` flag to the railway run command. This flag spins up a sidecar container (duplicating service), and the existing service is disconnected from it's upstream repo. The existing service will also automatically be deleted after 24h (TBD, not cli side).

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -3,7 +3,7 @@ use is_terminal::IsTerminal;
 
 use crate::{
     commands::ssh::common::{
-        create_spinner, establish_connection, execute_command, get_ssh_connect_params,
+        create_spinner, establish_connection, execute_command, get_ssh_connect_params, SSHArguments,
     },
     controllers::{
         environment::get_matched_environment,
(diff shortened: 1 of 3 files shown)


src/commands/run.rs
```diff
@@ -81,6 +89,86 @@ pub async fn command(args: Args, _json: bool) -> Result<()> {
...
+            let e = execute_command(&mut terminal_client, args.args.clone(), spinner).await;
+            if let Err(e) = e {
+                bail!("Failed to execute command: {e:?}")
+            } else {
+                Ok(())
+            }
+        });
+        // Now, we need to:
+        // 1. Duplicate the current service
+        // 2. Disconnect the current service (the one the command is running on) from upstream
```
coffee-cup, line 125: I think we want to disconnect the new service from upstream?

src/commands/ssh/common.rs
```diff
@@ -13,78 +13,47 @@ use crate::controllers::{
...
+///
+/// (project, service, environment, deployment instance id)
 pub async fn get_ssh_connect_params(
-    args: Args,
+    args: (
+        Option<String>,
+        Option<String>,
+        Option<String>,
+        Option<String>,
+    ),
```
coffee-cup, line 25: I think we should have a type/struct for these. Otherwise, it is easy to confuse what the correct order should be

src/commands/run.rs
Milo123459: Nope, this isn't what Angelo described

PR #623, opened 2025-04-28
Fix valkey "Template not found" in add command

lassejlv:
So when you are using "railway add" and choose valkey, you would get "Template not found"

So i replaced "valkey" with the template id "pQYeJx"
This does not happend with keydb beacuse it's template id is keydb.


paulocsanz (approved):

paulocsanz (changes requested):
Actually, I'm sorry, now that I got the whole context, we are not comfortable supporting new first class databases in the CLI, and merging the previous PR doing that was an oversight, we thought it would only affect the redis cli connect. Not railway add.

I will have to block this and revert this PR: https://github.com/railwayapp/cli/pull/620

We are happy to accept a PR passing the env vars to redis cli if that does not affect the other commands in railway cli.

PR #654, opened 2025-08-03
fix: update check no longer relies on config

alexng353:


coffee-cup (approved):
Looks good. In the future can you please add a PR description with reason and a point or two about the changes. It makes it much easier to review.

PR #668, opened 2025-09-25
Fix environment tokens not working

Milo123459:
- makes sure project access/environment tokens work in the ssh command (to ensure consistency)

note that specifying a token via RAILWAY_TOKEN and attempting to ssh still doesn't work.


mikewesthad (approved):
Dope, ty. Tested against latest backboard API changes Jared shipped and normal ssh + project token ssh work

Issue #635, opened 2025-06-09
Feature request: ability to set secret variables from CLI and read values from stdin


dctaf:
Railway CLI does not support setting secrets ("sealed" values) or reading values from stdin. When creating secrets, it can be convenient to generate them in a way that a human never sees them, e.g.

```
openssl rand -hex 16 | railway variables --sealed --set-from-stdin SECRET_KEY
```

PR #676, opened 2025-09-27
Add option to set from stdin using `railway variables`

jzeuzs:
Resolves https://github.com/railwayapp/cli/issues/635.


src/commands/variables.rs
```diff
@@ -74,6 +99,18 @@ pub async fn command(args: Args) -> Result<()> {
...
     };
 
+    if args.set_from_stdin {
+        let stdin = stdin();
+
+        for line in stdin.lock().lines() {
+            let line = line?;
+
+            if !line.trim().is_empty() {
+                args.set.push(line);
```
mikewesthad, line 109: Not super important b/c we trim later, but maybe this for consistency
```suggestion
                args.set.push(line.trim().to_string());
```

mikewesthad (commented):
Couple small things but otherwise LGTM. I think being able to do:

 ```
 echo "TEST=123" | cargo run variables
 ```
 
W/o the flag is a little cleaner, but I think what you have with an explicit flag is probably best here for backwards compat

jzeuzs:
> Couple small things but otherwise LGTM. I think being able to do:
> 
> ```
> echo "TEST=123" | cargo run variables
> ```
> 
> W/o the flag is a little cleaner, but I think what you have with an explicit flag is probably best here for backwards compat

I think having an explicit flag would be better.

ndneighbor:
Going to work on Milo's, and then yours after this one, but keeping an eye on this.

PR #679, opened 2025-10-08
ssh-command: Allow passing one or two of `-p`, `-e` and `-s`

malthejorgensen:
Previously, `railway ssh` required you to either be in a linked directory or pass all three `-p/--project`, `-e/--environment` and `-s/--service` arguments.

Now, if only one or two arguments are passed the ones that aren't passed will default to what was linked.

This allows more easily connecting to e.g. the production environment if you linked to the staging one, by simply passing:

    railway ssh -e <PRODUCTION_ENV_ID>

similarly if you want to connect to a service that wasn't linked to the current directory but is in the same project and environment you can now do:

    railway ssh -s <SERVICE_ID>

### DIFF
--- a/src/commands/ssh/common.rs
+++ b/src/commands/ssh/common.rs
@@ -1,10 +1,12 @@
 use std::io::Cursor;
 
-use anyhow::{anyhow, Result};
+use anyhow::bail;
+use anyhow::{anyhow, Context, Result};
 use indicatif::ProgressBar;
 use reqwest::Client;
 use std::io::Write;
 
+use crate::commands::queries::RailwayProject;
 use crate::controllers::{
     environment::get_matched_environment,
     project::get_project,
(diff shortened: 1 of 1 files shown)


malthejorgensen:
To really make the CLI as nice as Heroku's it would be great if the `--project` and `--service` args would support the project name and service name rather than just the IDs which have to be copied out manually from URLs or JSON.

For `--project` I believe this would require the GraphQL API to support looking up project by name (of course "404'ing" for project you don't have access to).

malthejorgensen:
Latest commit reverts arg-handling to how it was pre https://github.com/railwayapp/cli/pull/605 (unclear to me why arg-handling was changed as part of that). This allows environment and service to be passed as names rather than IDs.

coffee-cup (approved):
Looks good! thanks

PR #680, opened 2025-10-13
feat(service.rs): add service management features

anonrose:
This commit introduces significant enhancements to the service management capabilities of the application. It includes the ability to link a service to the current project, show the deployment status for services, and handle legacy direct service link. The changes also introduce new commands, arguments, and structures for managing services more effectively.

```bash
# Show status for linked service
railway service status

# Show status for a specific service
railway service status --service my-service

# Show status for all services in current environment
railway service status --all

# Show status in a different environment
railway service status --all --environment production

# Get JSON output
railway service status --json
```

### DIFF
--- a/src/commands/service.rs
+++ b/src/commands/service.rs
@@ -115,6 +115,7 @@ async fn link_command(args: LinkArgs) -> Result<()> {
 
         configs.link_service(service.0.id.clone())?;
         configs.write()?;
+        println!("Linked service {}", service.0.name.green());
         return Ok(());
     }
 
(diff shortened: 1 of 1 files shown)


coffee-cup (commented):
I would like to avoid a breaking change to the CLI. If just typing `railway service`, it should alias to the link command and prompt a service to link to.

coffee-cup (commented):
I'm also unsure about the subcommand structure of the service command. I could be wrong, but we don't currently have any subcommands like this so it would be unique and not follow the arg/flag based structure of the other commands.

coffee-cup (approved):
This is looking good! Just two comments
- Lints need to be fixed
- When linking to a service via the id (`service <serviceId>`), it no longer outputs anything. Can it be updated to log the service name and a success message

anonrose:
> This is looking good! Just two comments
> 
> * Lints need to be fixed
> * When linking to a service via the id (`service <serviceId>`), it no longer outputs anything. Can it be updated to log the service name and a success message

Will update this weekend, thanks!

PR #702, opened 2025-11-26
Allow changing service config in `railway environment new`

Milo123459:
This allows changing any service config non-interactively and interactively when creating a new environment, along with an existing environment via `railway environment edit`.

There is validation non-interactively that the dot-path provided exists (e.g doing something.arbitrary) would throw an error. If you are close to an option, it can suggest what option you may have meant.

It also performs type-based validation: based on the Rust struct `EnvironmentConfig` it will attempt to parse into that type, and if it fails, it will error.

# Examples

`railway environment new foo --duplicate bar --service-config baz source.repo "github/repo_link"`

`railway environment new foo --duplicate bar -s baz variables.KEY.value "hello, world!"` - add a new variable with key KEY and value "hello, world!"
It also works fully interactively:

https://github.com/user-attachments/assets/b165fefe-71a2-4436-aa27-be381ca5e8e5

It handles non-interactive errors as well:

image

It parses typos (e.g trailing & leading dots, or too many dots):

image

Distancing to give "did you mean" suggestions:

image

### DIFF
--- a/src/commands/environment/changes/source.rs
+++ b/src/commands/environment/changes/source.rs
@@ -16,35 +16,6 @@ pub enum SourceType {
     GitHub,
 }
 
-#[derive(Clone, Copy, Display, EnumIter, PartialEq)]
-pub enum AutoUpdateType {
-    #[strum(serialize = "Disabled")]
-    Disabled,
-    #[strum(serialize = "Patch versions only")]
-    Patch,
-    #[strum(serialize = "Minor versions")]
-    Minor,
-}
-
-impl AutoUpdateType {
-    fn to_api_value(self) -> &'static str {
-        match self {
-            AutoUpdateType::Disabled => "disabled",
-            AutoUpdateType::Patch => "patch",
-            AutoUpdateType::Minor => "minor",
-        }
-    }
-
-    fn from_api_value(value: &str) -> Option<Self> {
-        match value {
-            "disabled" => Some(Self::Disabled),
-            "patch" => Some(Self::Patch),
-            "minor" => Some(Self::Minor),
-            _ => None,
-        }
-    }
-}
-
 pub fn parse_interactive(
     service_id: &str,
     service_name: &str,
@@ -57,10 +28,6 @@ pub fn parse_interactive(
     let existing_branch = existing_source.and_then(|s| s.branch.as_deref());
     let existing_root_dir = existing_source.and_then(|s| s.root_directory.as_deref());
     let existing_check_suites = existing_source.and_then(|s| s.check_suites);
-    let existing_auto_update = existing_source
-        .and_then(|s| s.auto_updates.as_ref())
-        .and_then(|a| a.r#type.as_deref())
-        .and_then(AutoUpdateType::from_api_value);
 
     // Determine default source type based on existing config
     let default_source_index = if existing_image.is_some() {
@@ -106,23 +73,6 @@ pub fn parse_interactive(
             };
 
             entries.push((format!("{}.image", base_path), serde_json::json!(image)));
-
-            // Auto-updates (Docker only)
-            let auto_update_options: Vec<AutoUpdateType> = AutoUpdateType::iter().collect();
-            let auto_update_default = existing_auto_update
-                .and_then(|a| auto_update_options.iter().position(|o| *o == a))
-                .unwrap_or(0);
-
-            if let Some(auto_update) = prompt_options_skippable_with_default(
-                "Auto-update policy <esc to skip>",
-                auto_update_options,
-                auto_update_default,
-            )? {
-                entries.push((
-                    format!("{}.autoUpdates.type", base_path),
-                    serde_json::json!(auto_update.to_api_value()),
-                ));
-            }
         }
         SourceType::GitHub => {
             // GitHub repo (required)


Milo123459:
btw: currently, due to staged changes *not* being available on the public API, if you change the source & variables for a service it will make 2 deployments. if staged changes were available on the public API, this could be done in one deployment.

coffee-cup:
Staged changes should be public soon.

Thoughts on not prompting for a service to change the image source for by default? I'm worried about the complexity of this flow if we prompt for every possible thing that could change. Or instead we could have a "Do you want to configure a service" that goes into a subprompt to configure variables or a source image. It would allow us to extend what you can configure when creating a new environment in the future.

screenshot-2025-12-09-14 34 34

coffee-cup:
`railway environment edit` makes sense I think.

By any type of option, I mean all build/deploy/networking/source config. For reference, the staged change/environment config schema is available [here](https://backboard.railway.com/schema/environment.schema.json).

Milo123459:
Also- if applying an edit to an existing environment, and there are staged changes that aren't applied, interactively I'll prompt and show the changes and ask for confirmation to merge, but non-interactively should it merge the changes by default or not apply them at all, even the new ones?

I suppose I could also fetch the current changes, remove them, apply the new changes then apply the old changes again in a new patch and not stage those.

coffee-cup:
> note that the query for making a new environment seems to be slowing it down so much due to a change introduced waiting for the environment to be actually duplicated

There should be a flag for this so that you don't wait.

---

Taking a look at the commands via `--help`

```
railway environment new foo --duplicate bar --service-source <service name/service uuid> github nodejs/node/branch
```

How does this handle branches that have `/`s in them? We might want to make this a bit more explicit. Maybe `github nodejs/node,branch` instead? or `github nodejs/node branch`. or actually, we might even want this to be multiple flags?

Ideally the flags map directly to the actual schema that is available. This would mean that we don't have to update the CLI everytime a new config field is available. What about something like

```
railway environment new -d production \
  --service-config serviceId `source.repo` "nodejs/node" \
  --service-config serviceId `source.branch` "my-branch"
  --service-config serviceId `build.buildCommand` "my-build-command"
```

This would create a "patch" object that looks like

```json
{
  "source": { "repo": "nodejs/node", "branch": "my-branch" },
  "build": { "buildCommand": "my-build-command" }
}
```

<summary>
The condensed typescript type for the environment and patch schema is

<details>

```ts
interface EnvironmentConfig {
    services?: Record<string, {
      source?: {
        image?: string
        repo?: string
        branch?: string
        commitSha?: string
        rootDirectory?: string
        autoUpdates?: {
          type?: "disabled" | "patch" | "minor"
          schedule?: Array<{ day: number; startHour: number; endHour: number }>
        }
      }
      build?: {
        builder?: "NIXPACKS" | "DOCKERFILE" | "RAILPACK" | "HEROKU" | "PAKETO"
        watchPatterns?: string[]
        buildCommand?: string
        dockerfilePath?: string
        railpackVersion?: string
      }
      deploy?: {
        startCommand?: string
        preDeployCommand?: string[]
        numReplicas?: number
        healthcheckPath?: string
        healthcheckTimeout?: number
        sleepApplication?: boolean
        runtime?: "UNSPECIFIED" | "LEGACY" | "V2"
        registryCredentials?: { username: string; password: string }
        restartPolicyType?: "ON_FAILURE" | "ALWAYS" | "NEVER"
        restartPolicyMaxRetries?: number
        cronSchedule?: string
        region?: string
        multiRegionConfig?: Record<string, {
          numReplicas?: number
          stackerAssignment?: string
        }>
        overlapSeconds?: number
        drainingSeconds?: number
      }
      configFile?: string
    }>
  }
```
</details>
</summary>


This would also make the stdin input much easier too since the objects would be idential. We would just be generating an [EnvironmentConfig](https://github.com/railwayapp/cli/blob/master/src/controllers/config/environment.rs) object based on the user flags (which we already have a Rust struct for).

Does that make sense?

coffee-cup:
As for applying the change, you _should_ be able to apply the change while the environment is being created and we will apply the changes 1 by 1

Milo123459:
For the branches, I'll look into it. I have a feeling that the --service-config field is useful, but it does require the user to know a lot more about it - the current flags provide more of an explaination to the user just from the help command which is something I like. From a technical standpoint, the way each change works is that each has it's own non interactive parsing logic & interactive parsing logic, hence why each has it's own flag. But I can change it to --service-config, it would make a bit easier but maybe worsen UX. I still plan on adding the STDIN input though in JSON.

coffee-cup:
I don't think it will make the UX worth fwiw. We can still have examples in the `--help` output, which you would have to look at anyway to see what is possible.

src/commands/environment/changes/builder.rs
```diff
@@ -0,0 +1,133 @@
...
+                serde_json::json!("RAILPACK"),
+            ));
+            // Railpack doesn't have a config path option
+        }
+    }
+
+    // Watch patterns (applies to all builders)
+    let watch_placeholder = existing_watch_patterns
+        .map(|p| p.join(","))
+        .unwrap_or_else(|| "src/**,package.json".to_string());
```
coffee-cup: Why is watch patterns prompted for here? It should be a separate config option and not tied to the builder

coffee-cup: Why is watch patterns prompted for here? It should be a separate config option and not tied to the builder

src/commands/environment/changes/healthcheck.rs
```diff
@@ -0,0 +1,77 @@
...
+) -> Result<Vec<PatchEntry>> {
+    let existing_deploy = existing.and_then(|e| e.deploy.as_ref());
+    let existing_healthcheck_path = existing_deploy.and_then(|d| d.healthcheck_path.as_deref());
+    let existing_healthcheck_timeout = existing_deploy.and_then(|d| d.healthcheck_timeout);
+
+    let mut entries: Vec<PatchEntry> = Vec::new();
+    let base_path = format!("services.{}.deploy", service_id);
+
+    // Health check path
+    let path_placeholder = existing_healthcheck_path.unwrap_or("/health");
```
coffee-cup: The default should be null/None, not an actual value (same with other config fields that can be empty)

src/commands/environment/edit.rs
```diff
@@ -0,0 +1,284 @@
...
+    // Get commit message: from --message flag, or prompt interactively (only if input wasn't from stdin)
+    let commit_message = if let Some(msg) = args.message.clone() {
+        fake_select("Commit message", &msg);
+        Some(msg)
+    } else if is_interactive {
+        prompt_text_with_placeholder_disappear_skippable(
+            "Commit message <esc to skip>",
+            "optional",
+        )?
+        .filter(|s| !s.is_empty())
```
coffee-cup: If no commit message is provided, I don't think we should prompt. It is very very optional and adding the extra step is unnecessary friction.

src/controllers/config/patch.rs
```diff
@@ -0,0 +1,694 @@
...
+pub fn is_empty(config: &EnvironmentConfig) -> bool {
+    config.services.is_empty()
+        && config.shared_variables.is_empty()
+        && config.volumes.is_empty()
+        && config.buckets.is_empty()
+        && config.private_network_disabled.is_none()
+}
+
+#[cfg(test)]
+mod tests {
```
coffee-cup: Nice set of tests

src/commands/environment/edit.rs
```diff
@@ -0,0 +1,284 @@
...
+    let env_config = get_edit_config(
+        &args,
+        &client,
+        &configs,
+        &project,
+        &environment_id,
+        stdin_is_terminal,
+    )
+    .await?;
+
```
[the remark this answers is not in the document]
Milo123459: it's just a json serialized environment config that gets passed straight to the API

coffee-cup (approved):

PR #853, opened 2026-04-22
feature: add `railway version` command with build metadata

manwindersingh1712:
## Description
  Adds a new `railway version` command that displays detailed build information and update status.                                                                      
                                                                                                                                                                      
  ## Features                                                                                                                                                           
  - Shows CLI version, git commit hash, build date, Rust version, and target triple                                                                                   
  - Checks for available updates from GitHub releases
  - Displays upgrade instructions when an update is available
  - Supports `--json` flag for programmatic usage

  ## Example Output
>When you are on latest version
<img width="555" height="320" alt="Screenshot 2026-04-22 at 4 03 19 PM" src="https://github.com/user-attachments/assets/36f3c5bd-d271-4067-b669-c88bed41a30f" />

> Example 2 : When there is an update
<img width="591" height="330" alt="Screenshot 2026-04-22 at 4 05 35 PM" src="https://github.com/user-attachments/assets/285961b6-12fd-410e-9a37-c2d7af78eff0" />

  ## Changes
  - Added `src/commands/version.rs` - new command implementation
  - Modified `build.rs` - expose `GIT_SHA`, `BUILD_DATE`, and `RUSTC_VERSION`
  - Registered command in `src/commands/mod.rs` and `src/main.rs`

  ## Testing
  - [x] `cargo build --release` succeeds
  - [x] `railway version` shows correct output
  - [x] `railway version --json` outputs valid JSON
  - [x] Update check works (shows status based on latest release)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -5,8 +5,41 @@ fn main() {
     println!("cargo:rerun-if-changed=src/gql/subscriptions/strings");
     println!("cargo:rerun-if-changed=src/gql/schema.json");
 
+    // Rebuild when git HEAD changes
+    println!("cargo:rerun-if-changed=.git/HEAD");
+
     // Expose the compile-time target triple so the self-updater fetches the
     // correct release asset (respects ABI: gnu vs musl, msvc vs gnu, etc.).
     let target = std::env::var("TARGET").unwrap();
     println!("cargo:rustc-env=BUILD_TARGET={target}");
+
+    // Expose git commit hash
+    let git_sha = std::process::Command::new("git")
+        .args(["rev-parse", "--short", "HEAD"])
+        .output()
+        .ok()
+        .and_then(|output| String::from_utf8(output.stdout).ok())
+        .map(|s| s.trim().to_string())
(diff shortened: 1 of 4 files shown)


noahdunnagan:
I'm a bit confused as to why we would add `railway version` instead of just extending the existing `❯ railway --version`.

manwindersingh1712:
> I'm a bit confused as to why we would add `railway version` instead of just extending the existing `❯ railway --version`.

@noahdunnagan  My thinking here is that --version should stay simple and fast, it’s meant for scripts and piping. Changing it to do network calls or print multi-line output could break things like `railway --version | grep 4.40.`

Version is meant to handle the richer, diagnostic use case (commit, build date, etc.), which is more useful for debugging and CI.

This follows the pattern most CLIs use (Docker, kubectl, Terraform), so separating the two keeps things clean and reliable.

manwindersingh1712:
@noahdunnagan Something like this.
Screenshot 2026-04-22 at 5 38 35 PM

PR #900, opened 2026-05-13
Add volume file browser and transfer commands

ThallesP:
## Summary

Adds volume file browsing and agent-friendly file transfer commands under `railway volume`.

https://github.com/user-attachments/assets/5144e1bf-f274-42a7-bcc7-d0fb6be9a67f

- adds `railway volume browse` with a Ratatui remote browser, download/upload actions, and edit-and-sync for files
- adds non-interactive `volume ls`, `volume download`, and `volume upload` commands with JSON output support
- uses native SSH with a multiplexed control connection; file transfers use `cat`, directory transfers use streamed `tar`
- adds a local upload picker panel in the TUI instead of requiring typed upload paths

## Why

Railway volume files need a simple interactive workflow for humans and simple commands for automation/agents. The SSH control connection also avoids repeated SSH key passphrase prompts during a session.

## Validation

- `cargo fmt --check`
- `cargo check`

### DIFF
--- a/src/commands/volume.rs
+++ b/src/commands/volume.rs
@@ -3,9 +3,10 @@ use crate::{
     controllers::environment::get_matched_environment,
     controllers::project::{
         ProjectEnvironmentInstances, ProjectVolumeInstanceNode,
-        ensure_project_and_environment_exist, get_environment_instances, get_project,
-        volume_instances_in_env,
+        ensure_project_and_environment_exist, find_service_instance, get_environment_instances,
+        get_project, volume_instances_in_env,
     },
+    controllers::volume_browser::{VolumeBrowserParams, run as run_volume_browser},
     errors::RailwayError,
     queries::project::ProjectProject,
     util::{
@@ -17,7 +18,7 @@ use crate::{
 use anyhow::{anyhow, bail};
 use clap::Parser;
 use is_terminal::IsTerminal;
-use std::fmt::Display;
+use std::{fmt::Display, path::PathBuf};
 
 /// Manage project volumes
 #[derive(Parser)]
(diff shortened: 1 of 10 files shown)


src/commands/volume.rs
```diff
@@ -44,7 +49,7 @@ structstruck::strike! {
     #[strikethrough[derive(Parser)]]
     enum Commands {
         /// List volumes
-        #[clap(visible_alias = "ls")]
+        #[clap(visible_alias = "volumes")]
```
ThallesP: do not remove the alias, keep it intact.

we might need to rename ours from ls to something else though.

src/commands/volume.rs
```diff
@@ -128,6 +133,92 @@ structstruck::strike! {
...
+            volume: Option<String>,
+
+            /// Path to identity (private key) file to use, like `ssh -i`
+            #[clap(short = 'i', long = "identity-file", value_name = "PATH")]
+            identity_file: Option<PathBuf>,
+
+            /// Local directory used for downloads and relative uploads
+            #[clap(long, value_name = "PATH")]
+            local_dir: Option<PathBuf>,
+        }),
```
ThallesP, line 155: wondering if there's anything that I can do to prevent agents from using this (after all its a TUI and might confuse them). i could prob detect if it's agent via environment variables and tell it to use download/ls/upload.

maybe the is_terminal already help us with that? worth checking

src/commands/volume.rs
```diff
@@ -234,11 +325,485 @@ pub async fn command(args: Args) -> Result<()> {
...
+        service_instance_id: target.service_instance_id,
+        service_name: target.service_name,
+        volume_name: target.volume_name,
+        mount_path: target.mount_path,
+        local_dir: local_dir.unwrap_or(std::env::current_dir()?),
+        identity_file,
+    })?;
+
+    Ok(())
+}
```
ThallesP, line 437: we will need to check if the service is running, otherwise bail out (SSH command already does that btw).

same check for other commands related to this PR.

src/controllers/volume_browser/app.rs
ThallesP, line 1: would be nice if we supported vim mode (if we dont already).

basic navigation with hjkl is ok

src/controllers/volume_files.rs
ThallesP, line 1: we are using ssh with tar + cat, totally not ideal but something is wrong with SCP impl: https://discord.com/channels/713503345364697088/1501995980227088474.

i can migrate to SCP if that bug gets fixed.

PR #901, opened 2026-05-13
feat: Add support for SSH agents in addition to raw keyfiles

### LINKED ISSUE #870: `railway ssh` ignores `ssh-agent`, fails when keys aren't on disk
## Summary

`railway ssh` requires SSH keys to physically exist as `*.pub` files in
`~/.ssh/`. It does not consult `ssh-agent`, so any setup where keys are
agent-resident rather than on disk is broken — most notably:

- SSH'd into a remote dev host with `ForwardAgent yes` (private keys live
  on your laptop, the remote `~/.ssh/` is empty by design).
- 1Password SSH agent, `gpg-agent`, `keychain`, hardware tokens, etc.
- Any environment that intentionally keeps private key material off disk.

In all of these, `ssh somehost` works, `git push` works, but `railway ssh`
hard-fails before it ever calls `ssh`.

## Repro (agent forwarding — most common case)

1. From a laptop with keys loaded in `ssh-agent`, ssh into a remote host
   with agent forwarding enabled (`ForwardAgent yes` or `ssh -A`).
2. On the remote host:
   ```
   $ ls ~/.ssh/
   # empty (or just known_hosts)
   $ ssh-add -L
   ssh-ed25519 AAAA... laptop-key
   $ ssh git@github.com   # works fine via forwarded agent
   ```
3. `railway ssh` →
   ```
   No SSH keys found in ~/.ssh/

   Generate one with:
     ssh-keygen -t ed25519

   Then run this command again.
   ```

## Repro (1Password / agent-only laptop)

Same failure mode locally: 1Password's agent serves the key, `~/.ssh/` has
no `*.pub` files, `railway ssh` bails.

## Root cause

`src/controllers/ssh_keys.rs::find_local_ssh_keys()` enumerates keys by
`read_dir`-ing `~/.ssh/` and filtering by `.pub` extension. There is no
fallback to `ssh-add -L` / `SSH_AUTH_SOCK`.

`src/commands/ssh/native.rs::ensure_ssh_key()` hard-`bail!`s when that vec
is empty, even though `ssh` itself would happily authenticate via the
agent.

The same file also passes `-i <path>` to the spawned `ssh` in
`ensure_tmux_installed` / `run_tmux_session`. With an agent present, the
identity flag is unnecessary — `ssh` will negotiate via the agent.

## Suggested fix

1. In `find_local_ssh_keys()`, when `SSH_AUTH_SOCK` is set, also enumerate
   agent-resident keys via `ssh-add -L` and synthesize `LocalSshKey`
   entries (no `path`, just `public_key` + `fingerprint` + `key_type`).
2. In `ensure_ssh_key()`, treat agent keys the same as on-disk keys for
   the "is one of my keys already registered?" check.
3. In `ensure_tmux_installed` / `run_tmux_session`, skip `-i` when the
   selected key is agent-resident (or always skip `-i` when an agent is
   present and the agent advertises a matching fingerprint).

Happy to put up a PR if there's interest.

## Workaround

Export the public half of an agent-resident key to `~/.ssh/id_ed25519.pub`.
The private key is not needed on disk — `ssh -i <pubkey>` will use the
agent for the matching private key. But the CLI shouldn't require this,
and it's particularly painful on remote hosts where the user may not even
have the public key handy.

## Environment

- `railway --version`: 4.45.0
- OS: macOS 26.2
- Agent: 1Password SSH agent (also reproduces with forwarded OpenSSH agent)

KazWolfe:
Simple-ish attempt to fix #870 since it started to become a headache for me. I'll preface this PR by warning everyone that I am *not* a Rust developer, so I'm pretty sure a lot of things in here are wrong or deeply suboptimal (though I did have some people who are much better at Rust than me look at it and tell me to do arcane things). I'd be more than willing to bring in any fixes.

Changes:

* Read all SSH keys returned by `ssh-add -L` to query the agent, then read keys from `~/.ssh`.
  * If a key is in both locations, prefer the key from the agent.
* Add new methods `LocalSshKey#key_name` and `#key_source` to encapsulate some logic.
  * Key name is decided by comment, then file_stem, then hash.
  * Key source is the path or an SSH agent hint.
* Replace `hostname` with `comment` for semantic clarity.
  * Comments are now everything after the second split.

High level demo:

https://asciinema.org/a/wj2g7MpQf2AC82Wz

### DIFF
--- a/src/commands/ssh/keys.rs
+++ b/src/commands/ssh/keys.rs
@@ -279,7 +279,7 @@ async fn add_key(
     let local_keys = find_local_ssh_keys()?;
     if local_keys.is_empty() {
         bail!(
-            "No SSH keys found in ~/.ssh/\n\n\
+            "No SSH keys found in your SSH agent or ~/.ssh/\n\n\
             Generate one with:\n  ssh-keygen -t ed25519\n\n\
             Then run this command again."
         );
--- a/src/commands/ssh/keys.rs
+++ b/src/commands/ssh/keys.rs
@@ -251,7 +251,7 @@ async fn list_keys(workspace_id: Option<String>) -> Result<()> {
     if !unregistered.is_empty() {
         println!("Local Keys (not registered):");
         for key in unregistered {
-            let comment = key.key_comment.as_deref().unwrap_or(&"");
+            let comment = key.key_comment.as_deref().unwrap_or_default();
 
             println!("  {}", key.key_name());
             println!("    Fingerprint: {}", key.fingerprint);
--- a/src/commands/ssh/keys.rs
+++ b/src/commands/ssh/keys.rs
@@ -19,16 +19,7 @@ struct LocalKeyOption(LocalSshKey);
 
 impl fmt::Display for LocalKeyOption {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "{} ({})",
-            self.0
-                .path
-                .file_name()
-                .unwrap_or_default()
-                .to_string_lossy(),
-            self.0.fingerprint
-        )
+        write!(f, "{} ({})", self.0.key_name(), self.0.fingerprint)
     }
 }
 
@@ -176,18 +167,21 @@ async fn list_keys(workspace_id: Option<String>) -> Result<()> {
             // Extract comment/hostname from public key
             let parts: Vec<&str> = key.public_key.split_whitespace().collect();
             let key_type = parts.first().unwrap_or(&"");
-            let hostname = parts.get(2).unwrap_or(&"");
+            let comment = parts.get(2..).unwrap_or_default().join(" ");
 
             println!("  {}", key.name);
             println!("    Fingerprint: {}", key.fingerprint);
             if !key_type.is_empty() {
                 println!("    Type:        {}", key_type);
             }
-            if !hostname.is_empty() {
-                println!("    Hostname:    {}", hostname);
+            if !comment.is_empty() {
+                println!("    Comment:     {}", comment);
             }
             if local_match.is_some() {
-                println!("    Source:      local (~/.ssh/)");
+                println!(
+                    "    Source:      local ({})",
+                    local_match.unwrap().key_source()
+                );
             }
             println!();
         }
@@ -199,7 +193,7 @@ async fn list_keys(workspace_id: Option<String>) -> Result<()> {
         for key in &github_keys {
             let parts: Vec<&str> = key.key.split_whitespace().collect();
             let key_type = parts.first().unwrap_or(&"unknown");
-            let hostname = parts.get(2).unwrap_or(&"");
+            let comment = parts.get(2..).unwrap_or_default().join(" ");
             let fingerprint = compute_fingerprint_from_pubkey(&key.key).unwrap_or_default();
 
             // Check if already registered
@@ -210,8 +204,8 @@ async fn list_keys(workspace_id: Option<String>) -> Result<()> {
                 println!("    Fingerprint: {}", fingerprint);
             }
             println!("    Type:        {}", key_type);
-            if !hostname.is_empty() {
-                println!("    Hostname:    {}", hostname);
+            if !comment.is_empty() {
+                println!("    Comment:     {}", comment);
             }
             if is_registered {
                 println!("    Status:      registered");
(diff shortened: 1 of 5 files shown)


src/commands/ssh/keys.rs
```diff
@@ -257,20 +251,15 @@ async fn list_keys(workspace_id: Option<String>) -> Result<()> {
...
-                "  {}",
-                key.path.file_name().unwrap_or_default().to_string_lossy()
-            );
+            println!("  {}", key.key_name());
             println!("    Fingerprint: {}", key.fingerprint);
             println!("    Type:        {}", key.key_type);
-            if !hostname.is_empty() {
-                println!("    Hostname:    {}", hostname);
+            if !comment.is_empty() {
+                println!("    Comment:     {}", comment);
```
ndneighbor, line 260: Semantics, I would prefer if we didn't change this- such that we can keep things consistent. I would prefer that we add a separate method so that we don't overwrite existing behavior.

KazWolfe: Will fix the lint aside, however the semantics of `Hostname` are wrong (imo) here - SSH key comments are significantly more broad and cover things like card serial numbers, generating user/host (as probably implied), etc. I can revert if preferred, but the intent is to follow how SSH itself wants to work.

src/controllers/ssh_keys.rs
```diff
@@ -14,10 +15,29 @@ use crate::gql::queries::{GitHubSshKeys, SshPublicKeys, git_hub_ssh_keys, ssh_pu
...
+
+impl LocalSshKey {
+    pub fn key_name(&self) -> Cow<'_, str> {
+        match (self.key_comment.as_ref(), self.path.as_ref()) {
+            (Some(comment), _) if !comment.is_empty() => comment.into(),
+            (Some(_), None) => "SSH Agent Key".into(),
+            (_, Some(path)) => path.file_stem().unwrap().to_string_lossy(),
+            _ => (&self.fingerprint).into(),
+        }
+    }
```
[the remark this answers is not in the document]
KazWolfe: Fixed by setting `key_comment` to `None` if the split fails.

paulocsanz (approved):

src/commands/volume/ssh_key.rs
```diff
@@ -89,8 +89,12 @@ fn load_secret_key(path: &Path) -> Result<Result<russh::keys::PrivateKey, anyhow
 fn discover_private_key_paths() -> Result<Vec<PathBuf>> {
     let mut paths = Vec::new();
 
-    for public_key in find_local_ssh_keys()? {
-        if let Some(private_key_path) = private_key_path_for_public_key(&public_key.path) {
+    for public_key in find_ssh_key_files()? {
+        if public_key.path.is_none() {
```
KazWolfe: This was added in after I started working on this, so for now I just patched things such that this only works with SSH keyfiles. It'll take a bit more work in general to get agent support working properly, and it'd _really_ be better to refactor all of the key discovery mechanism(s) now that we have `russh` accessible to us.

(As an aside, this is a *very* fragile way of doing key discovery, but eh, it works.)

src/commands/volume/ssh_key.rs
```diff
@@ -89,8 +89,12 @@ fn load_secret_key(path: &Path) -> Result<Result<russh::keys::PrivateKey, anyhow
...
+            continue;
+        }
+
+        if let Some(private_key_path) = private_key_path_for_public_key(&public_key.path.unwrap()) {
```
paulocsanz: Lets avoid unwrapping here, use if let Some on public_key.path.as_ref() or something like that

src/controllers/ssh_keys.rs
```diff
@@ -59,17 +102,45 @@ pub fn find_local_ssh_keys() -> Result<Vec<LocalSshKey>> {
...
+
+            Ok(LocalSshKey {
+                path: None,
+                public_key: s.trim().to_string(),
+                fingerprint,
+                key_type: parts[0].to_string(),
+                key_comment,
+            })
+        })
+        .collect()
```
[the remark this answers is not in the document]
KazWolfe: This is pre-existing - any such explosion on file-based keys would do same. This is another good candidate for "rewrite the entire thing using russh." (which only became a thing *after* I opened this PR!)

KazWolfe:
A few questions for the Railway team (not specific to this PR, just in general for the `russh` changes):

1. OpenSSH by default only scans for `id_rsa`, `id_ecdsa`, `id_ecdsa_sk`, `id_ed25519`, and `id_ed25519_sk`. At present, Railway's SSH tooling scans for everything that matches `*.pub` and then attempts to find the base filename. This is technically a divergence from how OpenSSH tends to behave, which isn't necessarily a problem but is probably worth calling out. Would Railway prefer to continue using `.pub` scanning and then opportunistically attempting to load the private keys?
    1. There's a _slight_ argument to be made that it's better to scan every file and grab things that look like private keys instead. Then, either read from `.pub` or calculate the public components on the fly. Would this be desired instead?
2. Railway does not offer any support for [`-sk` FIDO2 credentials](https://cryptsus.com/blog/how-to-configure-openssh-with-yubikey-security-keys-u2f-otp-authentication-ed25519-sk-ecdsa-sk-on-ubuntu-18.04.html). Is this something that's desired?
3. At present, any invalid/failing SSH keys anywhere in the chain cause the entire process to abort early. Do we want to just skip invalid keys?

KazWolfe:
Closing in favor of #915.

PR #903, opened 2026-05-15
Support ssh-agent keys for native SSH

### LINKED ISSUE #870: `railway ssh` ignores `ssh-agent`, fails when keys aren't on disk

radiantjade:
## Summary
- include public keys from `ssh-add -L` when checking/registering SSH keys
- keep file-backed `~/.ssh/*.pub` keys preferred when the same fingerprint is also exposed by the agent
- update SSH key and DB stats messaging so agent-backed setups do not get told only to create files in `~/.ssh`

Fixes #870.

## Testing
- `PATH="$HOME/.cargo/bin:$PATH" cargo test`

### DIFF
--- a/src/commands/ssh/keys.rs
+++ b/src/commands/ssh/keys.rs
@@ -19,16 +19,7 @@ struct LocalKeyOption(LocalSshKey);
 
 impl fmt::Display for LocalKeyOption {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "{} ({})",
-            self.0
-                .path
-                .file_name()
-                .unwrap_or_default()
-                .to_string_lossy(),
-            self.0.fingerprint
-        )
+        write!(f, "{} ({})", self.0.display_name(), self.0.fingerprint)
     }
 }
 
@@ -187,7 +178,12 @@ async fn list_keys(workspace_id: Option<String>) -> Result<()> {
                 println!("    Hostname:    {}", hostname);
             }
             if local_match.is_some() {
-                println!("    Source:      local (~/.ssh/)");
+                println!(
+                    "    Source:      {}",
+                    local_match
+                        .map(|key| key.source_label())
+                        .unwrap_or_default()
+                );
             }
             println!();
         }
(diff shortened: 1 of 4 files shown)


radiantjade:
This PR will need one of the release labels for the label-check workflow. I do not have permission to add labels on this repo; release/patch should be appropriate for this fix.

Issue #922, opened 2026-05-29
Permission denied (publickey) when using railway ssh


bodgerbarnett:
I've used this command before - yesterday, in fact! But it's now telling me that my key is invalid.

❯ railway ssh --service API --environment production
Using SSH key from file <path to SSH key>
8811e921-9a1b-4131-bb10-135f76822c23@ssh.railway.com: Permission denied (publickey).

railway status tells me I'm logged in and my CLI is up to date as of today:-

❯ railway --version
railway 4.65.0

KazWolfe:
Hi! First off, my apologies, as this bug is *probably* related to my change(s) (ref #915). 

There's something a bit weird here, though. That log line is normally supposed to display the path of the file that was selected, which it isn't in this case.

Can you send the output of `railway ssh keys list` and the output of `ls ~/.ssh`? Alternatively, is your SSH key *not* in `~/.ssh`?

Lastly, do you know what version you were on prior to the upgrade? This specific question came up during the review process and we identified that this *shouldn't* happen, so I'm curious if you jumped a number of versions all at once. (see below the line for the technical details.)

---

T/N:

So, some quick background regarding how SSH keys are chosen. Ever since Railway rewrote the console connection feature to use true SSH, Railway itself did not choose the SSH key that would be presented to the server. Instead, Railway CLI would [simply invoke SSH directly](https://github.com/railwayapp/cli/blame/dc7142ab0e5a689dd82eced44463f45bee24c92e/src/commands/ssh/mod.rs#L117). That `identity_file` block would only have a value if `args` contained an identity via:

```rs
    /// Path to identity (private key) file to use, like `ssh -i`.
    /// Skips the local ~/.ssh scan; forwarded directly to ssh.
    #[clap(short = 'i', long = "identity-file", value_name = "PATH")]
    identity_file: Option<PathBuf>,
```

This `Option<PathBuf>` was then passed on to `run_tmux_session` or `run_native_ssh`, both of which still only [appends `-i` if `Some(key)` is defined](https://github.com/railwayapp/cli/blame/dc7142ab0e5a689dd82eced44463f45bee24c92e/src/commands/ssh/native.rs#L178). As best as I can tell, there's no way that (native) SSH would have ever picked up a key from `~/.ssh/*.pub` that wasn't named with the standard `id_` prefix, since that's all native SSH cares about unless it receives a `-i` hint.

[...]

PR #926, opened 2026-05-30
SSH Command: Handle Identity Files

KazWolfe:
Inspired by #925. If Railway wants to use a file-based SSH key, it will explicitly pass that key to the SSH command via `-i`. This helps cases where a user has a key at e.g., `railway_key`, which would not automatically be used by the SSH command. Should fix, e.g., #922.

Also fixes key sorting to always prioritize agent keys over file keys, as agents *generally* have tighter security guarantees.


m-abdelwahab:
LGTM. Thank you so much for the contribution @KazWolfe!

PR #936, opened 2026-06-04
fix(ssh): derive relay host/port from RAILWAY_ENV instead of hardcoding prod

codyde:
## Problem

The SSH relay host was a compile-time constant (`ssh.railway.com`) while backboard URLs follow `RAILWAY_ENV`. A dev-mode CLI therefore registers SSH keys and creates sandboxes against the **develop** backboard but dials the **production** relay — which has never seen the key or the target — yielding a confusing `Permission denied (publickey)` for a key the CLI just said was registered.

Reported by Pierre in raildev:
```
$ railway sandbox ssh
Using SSH key from agent: pborckmans@gmail.com
sbx:f597…:343d…@ssh.railway.com: Permission denied (publickey).
```
Manual workaround that proved the diagnosis: `ssh -p 2222 sbx:…@ssh.railway-develop.com` connects fine.

## Fix

`Configs::get_ssh_relay()` mirrors backboard's `controllers/ssh` mapping:

| env | relay |
|---|---|
| production | `ssh.railway.com` (port 22) |
| staging | `ssh.railway.com` — falls through like backboard's `IS_DEV`-only branch (`ssh.railway-staging.com` does not serve the SSH relay; probes accept TCP then close) |
| dev | `ssh.railway-develop.com` **-p 2222** |

Applied everywhere the constant was used: native ssh (`railway ssh`, `railway sandbox ssh`), tmux install/session, the generated `~/.ssh/config` block (gains a `Port` line only when non-default), and volume SFTP (russh connect tuple).

## Risk

Production behavior is byte-identical by construction: `("ssh.railway.com", None)` produces the same target string, adds zero ssh args, the same `(host, 22)` sftp tuple, and the same config block — the existing config-block tests asserting the literal prod output pass unmodified. The only behavioral change is dev mode, which previously failed 100% of the time.

## Testing

- Full test suite green (214)
- Prod smoke test on the built binary: `sandbox create` → `sandbox ssh -- echo` → `destroy`, all clean
- Dev-mode verification: Pierre has a reliable raildev repro; plain `railway sandbox ssh` should now connect exactly where his manual `-p 2222` command did

### DIFF
--- a/src/commands/sandbox.rs
+++ b/src/commands/sandbox.rs
@@ -551,7 +551,7 @@ async fn ssh(
     configs.write()?;
 
     // Relay target format (per backboard): sbx:<environmentId>:<sandboxId>.
-    // `run_native_ssh` appends `@ssh.railway.com` internally.
+    // `run_native_ssh` appends the environment's relay host internally.
     let target = format!("sbx:{environment_id}:{sandbox_id}");
 
     // Keep the sandbox alive for the duration of the session.
--- a/src/commands/ssh/config.rs
+++ b/src/commands/ssh/config.rs
@@ -272,8 +272,12 @@ fn render_config_block(
     writeln!(block, "# BEGIN railway:{rendered_marker}").expect("writing to String cannot fail");
     writeln!(block, "# Railway service: {rendered_service_name}")
         .expect("writing to String cannot fail");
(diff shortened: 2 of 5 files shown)


codyde:
Folding this into #933 — the fix originated on that branch and will ship with the sandbox release.

PR #953, opened 2026-06-08
feat(sandbox): `railway sandbox forward` — port-forward into a sandbox

codyde:
## What

Adds `railway sandbox forward`, which forwards local ports into a running sandbox — `kubectl port-forward` for Railway sandboxes. Run a dev server / database / API inside a sandbox and hit it from `localhost`.

```bash
railway sandbox forward 3000        # localhost:3000 → sandbox:3000
railway sandbox forward 8080:3000   # explicit local port (localhost:8080 → sandbox:3000)
railway sandbox forward 3000 5432   # several ports over one connection
railway sandbox forward 3000 --id <sandbox-id>
```

```
⚡ Forwarding to sandbox 2010daf4

  ➜  http://localhost:3000 → sandbox:3000

  Press Ctrl+C to stop
```

## Intended use

Ephemeral **"forward to this service for testing"** in the dev loop — not permanent public access. Public/shareable access for sandboxes is a separate, future effort (domains), which is a different mechanism (edge ingress, not a laptop tunnel). Scoping this command to the terminal workflow is deliberate.

## How it works

**No backend changes.** The tcp-proxy SSH relay already accepts SSH `direct-tcpip` channels for VM workloads, and sandboxes are microVMs — so the command just spawns:

```
ssh -N -L 127.0.0.1:<local>:127.0.0.1:<remote> sbx:<env>:<sandbox>@<relay>
```

All forwards multiplex over the session's single QUIC tunnel (measured ~32ms/request; 20 parallel connections in <100ms). It reuses the existing `sandbox ssh` key-registration flow and the sandbox heartbeat.

## Ergonomics

- Defaults to the **active sandbox**; ports are positional `[LOCAL:]REMOTE` specs
- Busy local port → bare specs fall back to a nearby free port (Vite-style) and say so; explicit `LOCAL:REMOTE` and `--strict` fail hard instead
- Multiple ports share **one** SSH connection
- Heartbeat keeps the sandbox alive while the forward is up
- Aliases: `port-forward`, `fwd`

## Robustness

- **First-run host key:** `-N` runs with stdin closed, so an interactive host-key prompt can't be answered — a fresh machine that hasn't trusted the relay would die with `Host key verification failed`. Set `StrictHostKeyChecking=accept-new` (TOFU; still rejects a *changed* key). The interactive `sandbox ssh` path doesn't need this since it inherits stdin.
- **Reconnect:** on an unexpected drop, re-check the sandbox; reconnect with exponential backoff while it's `RUNNING`, or report the real status (`STOPPED`/`GONE`) and exit when it actually stopped. A fast-failure budget prevents hot-looping on a broken relay/auth setup.

## Security note (for the relay, not this PR)

This command pins the remote side to `127.0.0.1`, matching the dashboard's `/ws/tcpip` bridge. Separately, the relay's native `direct-tcpip` handler passes the destination host through verbatim, so raw `ssh -L host:port` / `ssh -D` can egress *through* a sandbox today. That's pre-existing (any user with a registered key + sandbox can already do it) and not introduced here — but a blessed, documented forward feature raises how many people discover it. Worth a decision on whether the relay should pin loopback like the WS bridge does (~3-line tcp-proxy change) before GA. Not gating this PR.

## Testing

- Unit tests for `[LOCAL:]REMOTE` parsing (bare/explicit/garbage)
- Manual end-to-end against prod sandboxes: single port, busy-port remap, explicit-port-busy error, multi-port over one connection, first-run host-key (empty `known_hosts`), and sandbox-destroyed-mid-forward (reconnect loop reports real status).

### DIFF
--- a/src/commands/sandbox.rs
+++ b/src/commands/sandbox.rs
@@ -1513,8 +1513,7 @@ fn parse_port_spec(spec: &str) -> Result<PortSpec> {
 /// Bind-test then release — a small TOCTOU window before ssh re-binds, which
 /// is fine for the interactive dev workflow this serves.
 fn resolve_local_port(spec: &PortSpec, strict: bool) -> Result<(u16, bool)> {
-    let is_free =
-        |port: u16| std::net::TcpListener::bind(("127.0.0.1", port)).is_ok();
+    let is_free = |port: u16| std::net::TcpListener::bind(("127.0.0.1", port)).is_ok();
 
     let requested = spec.local.unwrap_or(spec.remote);
     if is_free(requested) {
@@ -1609,7 +1608,11 @@ async fn forward(
 
     let short_id: String = sandbox_id.chars().take(8).collect();
     eprintln!();
-    eprintln!("{} Forwarding to sandbox {}", "⚡".yellow(), short_id.bold());
+    eprintln!(
+        "{} Forwarding to sandbox {}",
+        "⚡".yellow(),
+        short_id.bold()
+    );
     eprintln!();
     for (remapped_remote, picked) in &remaps {
         eprintln!(
--- a/src/commands/sandbox.rs
+++ b/src/commands/sandbox.rs
@@ -7,7 +7,9 @@ use clap::Parser;
 use is_terminal::IsTerminal;
 
 use crate::client::{GQLClient, post_graphql};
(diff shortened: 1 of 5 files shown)


lawrencegripper (approved):
lgtm

PR #1005, opened 2026-07-09
Pin npm to 11.x for publish

futurepastori:
The publish-npm job runs on `setup-node` with Node 24, which bundles npm 12 — and npm 12's publish provenance path crashes with `Cannot find module 'sigstore'`. Adds an explicit `npm install -g npm@11` pin before publish; 11.5+ retains Trusted Publisher support. Unpin when the upstream issue closes.

Companion: railway-ts-sdk#49 pins its `npm@latest` trusted-publisher step the same way.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -230,6 +230,11 @@ jobs:
           node-version: 24
           registry-url: 'https://registry.npmjs.org'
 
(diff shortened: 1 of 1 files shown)


futurepastori:
Not needed — the exposure is only the SDK's explicit `npm install -g npm@latest` step; Node 24's bundled npm is 11.x. Pin lives in railway-ts-sdk#49.

PR #1026, opened 2026-07-25
feat(mcp): proxy the remote MCP server over stdio using the CLI login

codyde:
## What

Adds `railway mcp proxy`: a local stdio bridge to `mcp.railway.com` that attaches a fresh Bearer token from the stored `railway login` on every request. MCP-capable editors (Claude Code, Cursor, Codex, OpenCode, Copilot, Factory Droid) can use the remote MCP server **without the OAuth flow** (dynamic client registration + browser consent) and **without long-lived credentials in editor configs**.

Today an editor pointed at `mcp.railway.com` runs the full MCP OAuth dance. For developers who already ran `railway login`, that's a redundant second auth ceremony — and pasting a token into an editor config can't substitute for it because remote access tokens expire hourly. The proxy makes the existing CLI login *be* the auth.

## How it works

```
MCP client ──stdio/JSON-RPC──▶ railway mcp proxy ──HTTPS/Bearer──▶ mcp.railway.com
                                      │
                              ~/.railway/config.json
                          (auto-refreshed, lockfile-serialized)
```

- Token freshness delegates to `ensure_valid_token` (the same lockfile-serialized refresh the rest of the CLI uses), so the proxy coexists safely with the local `railway mcp` server and other CLI processes.
- SSE responses are re-framed as newline-delimited JSON-RPC through a single writer task (no interleaving); tool calls run concurrently after the handshake; transparent re-initialize + one retry on session loss.
- **Logged-out start heals without a restart**: the proxy completes the MCP handshake and returns an actionable "run `railway login`" on each call; the next request after login succeeds automatically.
- Bounded 5s shutdown so a stalled upstream can't orphan the process; JSON-RPC batch error paths answer every id.

## Install surface

`railway mcp install` gains a three-way transport:

- `--remote` → installs the **proxy** (`railway mcp proxy`), the new no-prompt default for CLI users.
- `--remote --oauth` → keeps the plain-URL, editor-managed OAuth flow (for environments without the CLI: web clients, hosted agents, teammates who never installed `railway`).
- no flag → local stdio server, unchanged.

Config matchers distinguish local / proxy / oauth entries (a bare "args contains `mcp`" check would misdetect a proxy install as local). Same three choices are plumbed through `railway setup agent`.

## Testing

- `cargo build` clean; `cargo test mcp` — 33 passing (SSE boundary/chunk-split parsing, stdout framing, batch handling, three-way matchers across JSON/opencode-array/codex-TOML config shapes).
- Verified end-to-end against `mcp.railway.com` through the built binary: initialize → 27 tools → live tool calls.
- Deep adversarial review pass; fixes for the logged-out heal, orphan-on-EOF, and batch error paths are included.

## Scope

MCP files only — deliberately excludes the unrelated `railway code` work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3775,6 +3775,7 @@ dependencies = [
  "base64",
  "bytes 1.11.1",
  "futures-core",
+ "futures-util",
  "http 1.2.0",
  "http-body",
  "http-body-util",
@@ -3795,12 +3796,14 @@ dependencies = [
  "sync_wrapper",
  "tokio",
  "tokio-rustls",
+ "tokio-util",
  "tower",
  "tower-http",
  "tower-service",
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
+ "wasm-streams",
  "web-sys",
  "webpki-roots",
 ]
@@ -5376,6 +5379,19 @@ dependencies = [
  "indexmap",
  "wasm-encoder",
  "wasmparser",
+]
+
+[[package]]
+name = "wasm-streams"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "15053d8d85c7eccdbefef60f06769760a563c7f0a9d6902a13d35c7800b0ad65"
+dependencies = [
+ "futures-util",
+ "js-sys",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "web-sys",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ reqwest = { version = "0.12.28", default-features = false, features = [
   "rustls-tls",
   "rustls-tls-native-roots",
   "json",
+  "stream",
 ] }
 reqwest-websocket = "0.5.1"
 regex = "1.12.3"
--- a/src/commands/mcp/install.rs
+++ b/src/commands/mcp/install.rs
@@ -14,20 +14,67 @@ pub struct Args {
     #[clap(long)]
     agent: Vec<String>,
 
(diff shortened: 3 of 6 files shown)


codyde:
## Security review (pre-merge)

Ran a dedicated threat-model pass on the proxy as a credential-handling component, since it injects a live OAuth Bearer on every request. Verified the redirect/header behavior against the actual reqwest 0.12.28 source, not from memory.

**Core is sound — no changes needed:**
- **Token can't be retargeted**: the destination URL is fixed at startup and never derived from server- or harness-supplied input; every POST/DELETE/re-init uses it verbatim.
- **Bearer is stripped on cross-host redirect**: confirmed in `reqwest/src/redirect.rs` (`remove_sensitive_headers` drops `AUTHORIZATION` when host/port differs), and it operates on the full outgoing header map including per-request headers.
- **TLS enforced** outside the Dev environment (`RAILWAY_ENV`-gated, not remotely flippable); **no log/error path stringifies** the token or the `Authorization` header; token isn't written anywhere new; harness message content can't inject request headers; no cross-session token-context races (single-user, token re-read per request).

**Hardened in `3416cc7`:**
1. **Bounded response buffers (32 MiB)** — reqwest's `text()`/`bytes()` are uncapped and the proxy sets no overall timeout, so a compromised edge / dev-MITM streaming a boundary-less SSE body could OOM a long-lived proxy. `stream_sse` + a new `read_body_capped` now abort past the ceiling.
2. **`redirect(Policy::none())`** — an MCP JSON-RPC POST should never redirect; closes the same-host 307/308 body-resend and redirected-response-relay edges outright.
3. **`RAILWAY_MCP_URL` validation** — reject non-`https` outside Dev so the override can't send the Bearer over plaintext (the first hop isn't covered by the cross-host strip).

34 tests pass; re-verified live streaming against `mcp.railway.com` (27 tools, tool calls stream fine) after the changes.

PR #1045, opened 2026-08-06
fix(environment): prioritize --service-config flags over auto-detected stdin in get_edit_config

### LINKED ISSUE #1044: `environment edit` silently ignores --service-config flags when stdin is not a TTY (exit 0, nothing saved)
## Summary

`railway environment edit --service-config <service> <path> <value>` silently discards the `--service-config` flags and exits `0` whenever **stdin is not a TTY** (scripts, CI, agent harnesses, or any `< /dev/null` / piped invocation). No change is committed, and in the default (non-`--json`, non-interactive) output mode nothing is printed to indicate the no-op — the command looks successful.

## Reproduction

CLI v5.28.0, macOS arm64 (code inspection shows the same logic on the v5.30.4 tag):

```console
$ railway environment edit --environment production \
    --service-config app deploy.preDeployCommand "npm run migrate" < /dev/null
> Environment production
$ echo $?
0
```

`railway environment config --environment production --json` afterwards shows no `preDeployCommand` — nothing was saved.

Adding `--json` reveals what happened:

```console
$ railway environment edit --environment production \
    --service-config app deploy.preDeployCommand "npm run migrate" --json < /dev/null
> Environment production
{"committed":false,"message":"No changes to apply","staged":false}
```

The exact same command run from an interactive terminal (stdin is a TTY) parses the flags and commits correctly.

## Root cause

`get_edit_config` in `src/commands/environment/edit.rs` gives piped stdin priority over explicitly passed CLI flags:

```rust
// Priority 1: Piped stdin JSON (auto-detected)
if !stdin_is_terminal {
    return read_config_from_stdin(environment_instances);
}

// Priority 2: CLI flags (--service-config, --service-variable)
if has_cli_flags {
    return parse_non_interactive_configs(&all_configs, environment_instances);
}
```

When stdin is non-TTY but empty, `read_config_from_stdin` returns `EnvironmentConfig::default()`, and the empty-config early-return only prints "No changes to apply" when `json` or `is_interactive` is true — so a scripted caller gets silence and exit code `0`.

This combination is easy to hit: any CI job or AI coding agent runs with non-TTY stdin, so the documented dot-path form no-ops there while appearing to succeed. It cost us a debugging session because the same command "worked on staging" (run interactively) and "failed on production" (run from a script) — the environment was never the variable.

## Expected behavior

Some combination of:

1. Explicit CLI flags should take priority over stdin auto-detection (check `has_cli_flags` first), or error loudly if both stdin content and flags are supplied.
2. When the resulting config is empty, print "No changes to apply" (stderr) in *all* output modes, and consider a non-zero exit code so scripts can detect the no-op.

## Environment

- CLI: 5.28.0 (reproduced); v5.30.4 source carries the same priority logic in `get_edit_config`
- OS: macOS arm64 (Darwin 25.3.0), zsh
- Workaround that behaves correctly in non-TTY contexts: pipe the JSON patch via stdin with service IDs, e.g. `railway environment edit --environment production --json <<'JSON' ... JSON`

AmariahAK:
### What was the issue?

`railway environment edit --service-config <SERVICE> <PATH> <VALUE>` silently discards the `--service-config` flags and exits `0` whenever stdin is not a TTY (scripts, CI, agent harnesses, or `< /dev/null`). No changes are committed, and in the default output mode nothing is printed — the command appears to succeed but does nothing.

### Where was the issue?

`src/commands/environment/edit.rs` — the `get_edit_config` function checked `!stdin_is_terminal` **before** `has_cli_flags`. When stdin was non-TTY (even empty), it short-circuited into `read_config_from_stdin`, which returned an empty `EnvironmentConfig::default()`. The explicitly passed CLI flags were never reached.

Additionally, the empty-config early-return only printed `"No changes to apply"` when `--json` or an interactive TTY was active. Non-TTY, non-JSON callers got silence and exit code `0`.

### How was it fixed?

Two changes in `get_edit_config`:

1. **Reordered the priority checks** — moved `has_cli_flags` (explicit `--service-config` / `--service-variable` flags) to Priority 1, above the `!stdin_is_terminal` auto-detection block. Explicit user intent always wins over environment detection.
2. **Added an `eprintln!` fallback** — when the resulting config is empty and neither `--json` nor an interactive TTY is active, `"No changes to apply"` now prints to stderr so scripted callers can detect the no-op.

### Why this approach?

- **Matches user intent:** someone passing `--service-config` meant to use it — the absence of a TTY shouldn't override that.
- **Matches existing codebase patterns:** the sibling function `edit_services_select` in `new.rs` already checks CLI flags first, then terminal state. `get_edit_config` was the outlier.
- **Minimal diff:** two blocks reordered, two lines added. No new functions, no new logic, no structural changes. Zero risk of breaking existing stdin JSON piping or interactive flows.

### How to test locally

```bash
# Build
cargo build

# Reproduction (should have been broken — flags discarded, no change saved)
echo "" | ./target/debug/railway environment edit \
    --environment production \
    --service-config <SERVICE> deploy.preDeployCommand "echo fixed" \
    --stage --json

# Verify the change was staged
railway environment config --environment production --json

# Regression: piping JSON via stdin must still work
echo '{"services":{"<service-id>":{"deploy":{"preDeployCommand":"from-stdin"}}}}' | \
    ./target/debug/railway environment edit --environment production --stage --json

# Regression: interactive mode must still work
./target/debug/railway environment edit --environment production
```

---

Fixes #1044

Original issue: https://github.com/railwayapp/cli/issues/1044

### DIFF
--- a/src/commands/environment/edit.rs
+++ b/src/commands/environment/edit.rs
@@ -69,6 +69,8 @@ pub async fn edit_environment(args: Args) -> Result<()> {
             );
         } else if is_interactive {
             println!("{}", "No changes to apply".yellow());
+        } else {
+            eprintln!("{}", "No changes to apply".yellow());
         }
         return Ok(());
     }
@@ -215,14 +217,15 @@ async fn get_edit_config(
     let all_configs = args.config.get_all_service_configs();
     let has_cli_flags = !all_configs.is_empty();
 
-    // Priority 1: Piped stdin JSON (auto-detected)
-    if !stdin_is_terminal {
-        return read_config_from_stdin(environment_instances);
-    }
-
-    // Priority 2: CLI flags (--service-config, --service-variable)
+    // Priority 1: CLI flags (--service-config, --service-variable)
+    // Explicit flags always take precedence over stdin auto-detection.
     if has_cli_flags {
         return parse_non_interactive_configs(&all_configs, environment_instances);
+    }
+
+    // Priority 2: Piped stdin JSON (auto-detected)
+    if !stdin_is_terminal {
+        return read_config_from_stdin(environment_instances);
     }
 
     // Priority 3: Interactive prompts (terminal only)


AmariahAK:
@codyde — could you add the release/patch label to this PR? I don't have permission as a fork contributor. Thanks!

PR #1058, opened 2026-08-08
fix: postgres pitr hardening — HA restore guard, pure JSON stdout, testable watch timeout

paulocsanz:
Follow-ups from the adversarial review + e2e coverage audit of the `railway postgres` stack (#1052–#1055).

## What

- **`pitr backup restore` refuses HA clusters.** An in-place restore replaces only the root's volume; replicas would rejoin with newer, diverged timelines. The CLI cannot reseed replicas (that path is dashboard-only for now), so it bails with a pointer to the dashboard.
- **`pitr enable --json` keeps stdout pure JSON** when the service is already enabled — the human-readable note goes to stderr, so `--json | jq` works on every path.
- **Debug-only env overrides** `RAILWAY_POSTGRES_WATCH_TIMEOUT_SECS` / `RAILWAY_POSTGRES_POLL_INTERVAL_SECS` (compiled out of release builds, same pattern as `RAILWAY_BACKBOARD_URL`) so the e2e harness can exercise the `progress --watch` timeout branch in seconds instead of 10 minutes.
- **`schema.json` regenerated** from the Public subgraph after railwayapp/mono#34881 exposed `replicaServiceIds`/`wipeServiceIds` on `volumeInstanceBackupRestore`. Semantic diff verified: those two args are the only structural change (the rest of the 36k-line diff is generated-file reflow).

## Tests

- 926/926 unit tests pass, `cargo fmt --check` clean.
- Hermetic mock e2e flows covering the new guard/branches land in the harness alongside this PR.


paulocsanz:
Added a third fix the new e2e step caught live: `pitr backup restore` only started the server-side copy — the volume swap stays as **staged changes** (dashboard flow applies them manually), so the CLI restore never actually took effect. The verb now waits for the copy, commits the staged swap (with the pre-existing-staged-changes warning), and deploys — "Restored … and deployed", consistent with the other verbs.

PR #1072, opened 2026-08-10
railway ca: workspace-first onboarding + railway agent harness

codyde:
## Summary
- Reworks the `railway ca` / `railway ca setup` first-run wizard's target step: it now shows workspaces first (the first one auto-expanded, mirroring the Manage tree), and expanding a workspace or project reveals what's under it — projects, then environments — instead of the old flat "pick from every project across every workspace" list.
- "+ Create a project" is now a row nested under the workspace you're browsing, and actually creates the project there (`Effect::CreateDefaultProject` now carries a `workspace_id`), instead of always silently defaulting to `workspaces.first()`.
- `railway ca setup` (the plain, non-TUI `inquire` flow) was audited against the same ask — its "Use an existing project"/"Create a project" paths already resolve workspace → project → environment in that order (`prompt_workspace_project_env`, `pick_workspace`), so no change was needed there.
- Adds `railway` as a fourth launchable harness alongside `claude`/`codex`/`grok` (`railway ca --railway` / `railway code --railway`). It's Railway's own agent (`railway-agent`/`railway-agent-tui`, built on pi-rs) — the VM already carries a server-minted LLM gateway credential and Railway platform tools at create time, so unlike the other three harnesses it needs no local sign-in copied or minted onto the agent (`PendingAuth::None`).

## Test plan
- [x] `cargo build` — clean
- [x] `cargo fmt -- --check` — clean
- [x] `cargo clippy --bin railway` — no new warnings in touched files
- [x] `cargo test --bin railway` — 965 passed (3 pre-existing, unrelated `auth_sim` failures reproduce identically on `master`, caused by this dev VM's own ambient Railway auth env leaking into those tests)
- [x] Downloaded the public `railway-agent`/`railway-agent-tui` v0.1.9 release binaries and ran them on this cloud-agent VM using its own ambient `AI_GATEWAY_URL`/`AI_AGENT_KEY` — confirmed a real one-shot `railway-agent run` call succeeds with zero local credential setup, validating the "own integrated credentials" premise this harness choice relies on.

### DIFF
--- a/src/commands/cloud_agent/lifecycle.rs
+++ b/src/commands/cloud_agent/lifecycle.rs
@@ -123,9 +123,9 @@ pub struct SshArgs {
     #[clap(long, value_name = "NAME")]
     session: Option<String>,
 
-    /// Leave the agent running on disconnect instead of putting it to sleep.
-    /// A running agent keeps billing for compute
-    #[clap(long)]
+    /// Accepted for compatibility; agents now always stay running on
+    /// disconnect. `railway ca sleep` stops the compute bill
+    #[clap(long, hide = true)]
     keep_awake: bool,
 
     #[clap(flatten)]
@@ -555,7 +555,14 @@ async fn ssh_connect(args: SshArgs) -> Result<i32> {
         telemetry::track_lifecycle("ssh_command", Duration::ZERO, None).await;
         run_command(&agent, &args.command).await
     } else {
-        attach(client, &backboard, &agent, args.session.as_deref()).await
+        attach(
+            client,
+            &backboard,
+            &agent,
+            args.session.as_deref(),
+            was_running,
+        )
+        .await
     };
 
     // A connection that never happened still woke a machine with no idle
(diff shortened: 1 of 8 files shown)


codyde:
Added a second commit: fixes `railway-agent-tui: terminal setup failed: The cursor position could not be read within a normal duration` when launching the railway harness through the `railway ca` TUI (as opposed to `railway code --railway`, which hands through a real terminal and was unaffected).

Root cause: the TUI's embedded session pane is a one-way `vt100` emulator — it renders whatever the remote side writes but never answers anything. `railway-agent-tui` sends a standard cursor-position probe (`ESC[6n`) during terminal setup and blocks on a reply; nothing was there to answer it, so it timed out. The reader thread now recognizes that query in the byte stream and writes back the emulator's own cursor position, same as a real terminal would.

Verified against the real `railway-agent-tui` v0.1.9 binary under a pty harness that mirrors `Session::spawn`: reproduces the hang without the fix, resolves cleanly with it.

codyde:
Added a commit ([`6ad1395`](https://github.com/railwayapp/cli/commit/6ad1395)) that stops sleeping agents on disconnect and defends against zombie-session reattach.

**Why here:** while testing this PR set locally, reattaching to listed sessions showed blank panes. Root cause: sleep kills every process on the VM, but `cloudAgentConsoleSessions` keeps reporting those sessions as `runState.running: true` (verified: `ps -ef` on a woken agent shows only kernel threads while the API lists a running session). Attach-by-name to a dead session streams zero bytes, forever. Every automatic sleep-on-disconnect was manufacturing these zombies.

**Changes:**
- `railway code`, `railway ca ssh`, and TUI quit now detach only — agents stay running. Sleeping is deliberate (`railway ca sleep`, or `s` on the tree). `--keep-awake` stays accepted (hidden) for compatibility; the failed-connect rollback sleep stays since it undoes a wake that never delivered a session.
- `ca ssh` attach: an agent that was asleep moments ago can't have a live session, so claimed-running sessions are skipped with a note and a fresh one starts.
- TUI reattach pane: an attach that delivers no bytes within 5s now says "the session may have ended when the agent slept" instead of staying silently blank.

The platform-side fix (invalidating session runState on sleep) is still worth doing — this makes the CLI honest in the meantime.

PR #1073, opened 2026-08-10
Ask which workspace the wizard's new project lands in

codyde:
First-run setup's "Create a project" made "Cloud Agents" in
`workspaces.first()` and said nothing about it. On an account with more
than one workspace that is a coin flip, and the project is then invisible
to anyone who cannot see the workspace it guessed — with nothing in the
flow to say which one that was. The CLI's own `railway ca setup` has
always prompted here (`pick_workspace`); only the TUI wizard guessed.

Add the question it was missing: choosing "Create a project" with more
than one workspace now goes to a picker first, listing each workspace and
how many projects it holds. One workspace still creates straight away —
there is nothing to ask. The workspace rides through the create effect,
and an id that no longer resolves is an error rather than a quiet
fallback to the first, which is the same bug wearing a different hat.

Where it went is now said twice over: the busy line names the workspace
while the project is being created, and the status line reports it
afterwards. The workspace question shares the project step's progress
dot, since it is the same question — where agents live.

WorkspaceNode carries its id for this; the tree grouped by name alone.

### DIFF
--- a/src/commands/cloud_agent/mod.rs
+++ b/src/commands/cloud_agent/mod.rs
@@ -73,7 +73,9 @@ async fn browse() -> Result<()> {
     let tree = tui::load_tree(&client, &configs).await;
     spinner.finish_and_clear();
     let tree = tree?;
-    if tree.is_empty() {
+    // Workspaces alone are not enough to browse — the listing now keeps the
+    // empty ones, for setup to create a project in — so ask about projects.
+    if tree.iter().all(|ws| ws.projects.is_empty()) {
         println!(
             "No projects with environments you can use. Create one with {} first.",
             "railway init".cyan()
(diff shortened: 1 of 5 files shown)


codyde:
Closing as superseded by #1072, which fixes the same bug — a new project silently landing in `workspaces.first()` — with a different design.

Where this PR adds a `WorkspacePick` step in front of the create, #1072 replaces the flat project list with an expandable workspace → project → environment tree and nests "+ Create a project" *under* the workspace you are browsing, so the workspace is answered by where you are rather than by an extra question. Both rewrite `Wizard::new`'s signature and both change `Action::CreateProject` to carry a workspace, so they cannot both land — they conflicted across all four wizard files.

The intent here isn't lost. #1072 carries the workspace through the create effect and errors on an id that no longer resolves rather than falling back to the first, and the settings card in #1069 now offers one create row per workspace for the same reason.

PR #1094, opened 2026-08-12
feat(mcp): default new agent installs to the remote CLI proxy

codyde:
## Summary
- New agent installs now default to the **remote CLI proxy** (`railway mcp proxy` → `mcp.railway.com`, auth via `railway login`):
  - `railway setup agent -y` / non-TTY setup
  - `railway mcp install` (no flags)
  - `curl -fsSL agents.railway.com | sh` / `install.sh --agents`
- Interactive setup picker lists **Remote** first as the default option.
- Escape hatches:
  - `--local` → old in-process `railway mcp`
  - `--oauth` → plain `https://mcp.railway.com` (editor OAuth)
  - `--remote` kept as a compatibility alias of the new default
- **Re-run behavior:** if a user already has local MCP configured and re-runs `setup agent` / `mcp install`, their railway entry is updated to the proxy. That migration-on-rerun is intentional for this phase.
- **Not in this PR:** bare `railway mcp` still runs the local server. Users who never re-run setup/install keep local until the later runtime default flip.

## Why
Canary the remote-proxy path on new installs and voluntary re-runs before flipping bare `railway mcp` for everyone.

## Test plan
- [x] `cargo test commands::setup::tests`
- [x] `cargo test from_flags_defaults_to_remote_proxy`
- [x] `cargo test mcp_install_subcommand`
- [x] `cargo test setup_agent_subcommand`
- [x] Fresh `$HOME`: `railway setup agent -y` writes `args: ["mcp","proxy"]`
- [x] Existing local entry + `mcp install` / `setup agent -y` migrates to proxy
- [x] `railway mcp install --local` still writes local `args: ["mcp"]`
- [x] `--local` conflicts with `--remote` / `--oauth`
- [ ] Smoke agents.sh / install.sh --agents against a current CLI binary

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,13 +61,21 @@ Configure Railway agent support for AI coding tools:
 railway setup agent -y
 ```
 
-This installs Railway skills and configures the Railway MCP server for detected tools such as Claude Code, Cursor, Codex, OpenCode, GitHub Copilot, and Factory Droid.
+This installs Railway skills and configures the remote Railway MCP server (`mcp.railway.com` via `railway mcp proxy`, authenticated with your CLI login) for detected tools such as Claude Code, Cursor, Codex, OpenCode, GitHub Copilot, and Factory Droid.
 
 Use the focused commands when you only need one part of the setup:
 
 ```bash
 railway mcp install --agent cursor
 railway skills --agent claude-code
+```
+
+Opt into the local GraphQL-backed MCP server instead:
+
+```bash
+railway setup agent -y --local
+# or
+railway mcp install --local --agent cursor
 ```
 
 ## Contributing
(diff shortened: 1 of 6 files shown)


codyde:
Follow-up commit addresses the open pressure-test items on the CLI side:

1. **Installer version skew** — non-interactive `install.sh --agents` now probes `setup agent --help` and passes `--remote` when supported (unless `--local`).
2. **Flag semantics** — help text clarifies `--oauth` is editor OAuth (not proxy), `--remote` is a default alias and ignored with `--oauth`, with a runtime warning if both are passed.
3. **Interactive default** — MCP transport picker starts on Remote (index 0).

Mono `agents.md` docs lag remains a separate follow-up.

PR #1123, opened 2026-08-24
Stick config pull/init to the existing authoring language

futurepastori:
## Summary
- If `.railway/railway.py` or `railway.go` already exists, `railway config pull` / `init` rewrite that file instead of always emitting TypeScript.
- Language is never inferred from `package.json` / `go.mod` / the app repo. No file yet still means TypeScript. Two files at once is an error.
- Pull printers for Python and Go share the same graph → source path the TypeScript importer already used in the CLI.

## Test plan
- [x] `cargo test --bin railway commands::config::`
- [ ] `railway config pull` in a directory with only `.railway/railway.py` updates the Python file
- [ ] `railway config pull` in a directory with only `.railway/railway.go` updates the Go file
- [ ] two authoring files together errors
- [ ] no authoring file still writes `.railway/railway.ts`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/commands/config/authoring.rs
+++ b/src/commands/config/authoring.rs
@@ -1,0 +1,76 @@
+use std::path::Path;
+
+/// Authoring language for `.railway/railway.{ts,py,go}`.
+///
+/// Never inferred from the application repo (`package.json`, `go.mod`, …).
+/// Once one of those files exists, every command that writes config must use it.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


futurepastori:
Holding merge until `railway-iac` (PyPI) and `github.com/railwayapp/railway-go-iac` (tagged module) are published.

Pulled Python/Go files import those packages; `railway config plan` cannot evaluate them until install works for users.

PR #1127, opened 2026-08-25
fix(tui): make copy to clipboard work on Linux (Wayland and X11)

lawrencegripper:
Copy in the cloud-agent and develop TUIs silently failed on Linux: arboard was built without Wayland support, and on X11 the `Clipboard` was dropped right after `set_text`, releasing the selection before anything could paste it.

Enables arboard's `wayland-data-control` feature and routes the three call sites through a shared `util::clipboard::copy` that holds one clipboard handle for the life of the process.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,6 +192,7 @@ dependencies = [
  "parking_lot",
  "percent-encoding",
  "windows-sys 0.60.2",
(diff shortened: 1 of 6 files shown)


codyde:
Checked this out, its solid

PR #1133, opened 2026-08-26
fix(postgres): honest switchover failures — SSH preflight and per-member probe errors

paulocsanz:
**The misleading message**: `Could not reach any cluster member's Patroni API to determine the current leader.` has been conflating a wedged cluster with the CALLER's own SSH setup. `railway postgres ha switchover` drives Patroni through `exec_in_container` — native `ssh <instanceId>@ssh.railway.com` — which needs a registered key and does no preflighting (`exec.rs` says so explicitly). An API token alone reaches nothing, so every headless runner / CI / user without a registered key gets the cluster-sounding error while the cluster is perfectly healthy.

**Live case**: the postgres-cli e2e battery hit exactly this on every switchover for weeks (16 retries/tick, cluster healthy, leader elected) until its runner got a registered key — details in paulocsanz/test-postgres-cli#6.

Two changes:
- **SSH preflight in switchover**: `ensure_ssh_key_quiet` on interactive runs (may register a key on the spot, as elsewhere in the CLI), `ensure_ssh_key_noninteractive` under `--yes` (never blocks on a prompt). Failure says what's actually wrong and carries the existing recipe (`ssh-keygen -t ed25519` + rerun).
- **`probe_any` returns per-member errors** on total failure and the switchover bail prints them — `SSH command failed (exit code 255): Permission denied (publickey)` vs `Timed out probing Patroni` finally distinguish a local auth problem, a dead member, and a wedged cluster.

`cargo test`: 1217 passed. `cargo check` clean.

Note: `ha status`/`pitr status` live probes share the same SSH path and still degrade silently to "unknown" — a follow-up could add a one-line stderr note there; kept out of this PR to stay surgical on the switchover contract.

### DIFF
--- a/src/commands/postgres/ha.rs
+++ b/src/commands/postgres/ha.rs
@@ -763,10 +763,37 @@ async fn switchover(
     .await
     .context("Failed to resolve live cluster member instances")?;
 
+    // The probes below run over native `ssh <instance>@ssh.railway.com` —
+    // an API token alone reaches nothing. Preflight the key so the failure
+    // is "your SSH setup" (with the recipe) instead of a misleading
+    // "could not reach the cluster". Interactive runs may register a key on
+    // the spot; --yes runs must never block on a prompt.
+    let preflight = if args.yes {
+        crate::commands::ssh::native::ensure_ssh_key_noninteractive(&ctx.client, &ctx.configs).await
+    } else {
+        crate::commands::ssh::native::ensure_ssh_key_quiet(&ctx.client, &ctx.configs).await
+    };
+    if let Err(e) = preflight {
+        bail!(
+            "Switchover drives Patroni over SSH (ssh <instance>@ssh.railway.com), and no usable \
+             SSH key is available: {e:#}"
+        );
+    }
+
     let probe_targets: Vec<String> = instance_ids.values().cloned().collect();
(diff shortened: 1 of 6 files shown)


paulocsanz:
Scope extended per follow-up ("retry everything that uses probes on intermittent failure"), second commit `51e5cd5`:

- **`exec_probe_in_container`**: every live READ probe (Patroni `/cluster`, `ha status` liveRole, pgBackRest info, archiver query, PgBouncer admin console, `railway metrics` DB stats) now retries transient transport failures — 3 attempts, 1s/3s pauses, per-attempt timeout INSIDE the loop, and a transient/deterministic classifier (auth failures, missing binaries, host-key refusals fail immediately; unit-tested table).
- **Mutations never retry**: Patroni's `POST /switchover` stays single-shot on the raw exec — replaying after a timeout that actually landed would double-switch.
- **PITR probes get a per-attempt budget** (previously unbounded — a hung relay stalled `pitr status`), and `follow_progress`'s GraphQL poll rides out up to 5 consecutive `FetchError`s within its own deadline (folds in the in-tree WIP that did exactly this — it was complete, compiling and green, so it ships instead of sitting uncommitted).

`cargo test`: 1218 passed.

PR #1144, opened 2026-08-27
feat(database): add MySQL point-in-time recovery

paulocsanz:
Stacked: **2 of 2**, on top of #1143 — review that one first. **Base is `pcs/mysql-redis-ha-cli`**, so the diff here is only the PITR delta (4 files, ~180 lines).

## What this adds

`railway mysql pitr` — binlog archiving into a Railway bucket, enabled by the `mysql-pitr` composable overlay.

The generic `pitr` tree (from #1143) already drives everything an engine's archive needs from its declared contract, so this is mostly the declaration plus the two ways MySQL genuinely differs from Postgres. Both become declarations rather than branches on the engine name:

**`supports_ha: false`** — the image's archiver refuses to run whenever the Group Replication seed list is set, so MySQL PITR is standalone-only. Every path that would take or follow the rolling HA workflow checks this first:

- `progress` / `cancel` / `clear` refuse **before any network call**
- `enable` / `disable` refuse **before the HA mutation**

so you get one clear sentence instead of a server error from a workflow that was never going to start:

```
$ railway mysql pitr progress
MySQL PITR is standalone-only: its archiver does not run on HA cluster
members, so there is no rolling PITR workflow progress to show.
```

**`probe_kind: None`** — the live coverage probe in `pitr status` is pgBackRest shelling into the container, which MySQL has no equivalent tool for. Selecting the probe by declared kind means MySQL renders no coverage section at all, rather than a probe run with another engine's tooling or a fake "unavailable" for something it simply doesn't implement.

## What needed no code

- **The archive variable contract.** `BINLOG_ARCHIVE_` as the declared prefix is all the enabled-state detection and the enable overlay need.
- **Image eligibility.** It ships with the `mysql-pitr` template record. Worth noting that template does **not** set `requireFloatingMajorTag`, unlike `postgres-pitr` — its image matrix only publishes exact minors, so the old hardcoded "minor pins are bad" rule would have refused every MySQL image. Reading the declaration is what makes both engines correct at once.
- **Restore, backups and schedules.** These ride the volume-instance mutations, which are engine-agnostic server-side.

## Verification

- `cargo test` — 1305 pass. New coverage for the per-engine archive-variable contrast (a Postgres service reads as unconfigured under MySQL's contract and vice versa), the HA-workflow gate refusing MySQL while passing Postgres, and probe-kind selection.
- `cargo clippy` / `cargo fmt` clean.
- The `mysql-pitr` template record was read from the live API; the fixtures mirror it.

### DIFF
--- a/src/commands/database/pitr.rs
+++ b/src/commands/database/pitr.rs
@@ -1,14 +1,18 @@
-//! The `pitr` verb -- point-in-time recovery / continuous backups, for every
-//! engine whose registry entry declares a PITR contract.
+//! `railway {postgres,mysql} pitr` -- point-in-time recovery / continuous
+//! backups, for every engine whose registry entry declares a PITR contract.
 //!
-//! The engine declaration decides which composable template the enable
-//! overlay deploys and which variables carry the archive
-//! (`PitrSpec::template_code`, `archive_var_prefix`). Which images may adopt
-//! the overlay is not decided here at all: that rule ships with the enable
-//! template (`adoptionImageEligibility`) and is read off the fetched record,
-//! so widening it is a template update rather than a CLI release. Backups,
+//! The engine declaration decides three things here: which composable
+//! template the enable overlay deploys (`PitrSpec::template_code`), whether
+//! the rolling HA enable/disable workflow exists at all (`supports_ha` --
+//! MySQL's archiver only runs standalone, so its HA path is a refusal, not a
+//! workflow), and which live coverage probe backs `status` (`probe_kind` --
+//! only pgBackRest ships one). Which images may adopt the overlay is not
+//! decided here at all: that rule ships with the enable template
+//! (`adoptionImageEligibility`) and is read off the fetched record, so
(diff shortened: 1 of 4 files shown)


paulocsanz:
Note on the thin check list: `ci.yml` triggers only on `pull_request: branches: [master]`, so a PR based on a feature branch runs `label-check` and nothing else. The full suite will run here once #1143 merges and GitHub retargets this to `master`.

Verified locally on this exact commit in the meantime:

```
cargo test    1286 passed; 0 failed
cargo clippy  no new warnings in any touched file
cargo fmt     clean
```

paulocsanz:
Rebased onto v5.47.0 (`c258497`, replacing `179ddfa`): the old base branch `pcs/mysql-redis-ha-cli` predated #1149's commit-staged wait and #1159's five fixes, both now on master. The replay was conflict-free and the PR diff now shows only the MySQL PITR commit. `cargo test` 1312 pass, clippy/fmt clean on the touched files.

PR #1151, opened 2026-08-28
feat(cloud-agent): report the pane's session to herdr

itsRems:
Stacked on #1150.

# What

The local half of exact-session restore for terminal workspace managers: when `railway code` / `railway ca ssh` runs inside a [herdr](https://herdr.dev) pane, the CLI reports the active session's `railway code --resume <agent>:<session>` reference over the pane's socket, the same protocol herdr's own agent integrations use. A workspace manager that stores the reference can put the pane back into the exact cloud-agent conversation after a restart — the way it already does for local agents.

Reporting is gated on herdr's pane contract (`HERDR_ENV=1` + `HERDR_SOCKET_PATH` + `HERDR_PANE_ID`), entirely best-effort, and silent: outside a herdr pane it is a no-op, and a dead or slow socket can never fail a session.

# Changes

- New `cloud_agent::herdr` module: pane-contract detection (read once), the framed `pane.report_agent_session` request (agent `railway-code`, source `railway:cli`), and a fire-and-forget unix-socket send with 500ms timeouts. Unix-only transport; other platforms no-op.
- The manage TUI reports the active pane's session whenever it changes — a launch landing, `adopt_pane_sessions` renaming a minted name to the relay-listed one, the cursor moving between sessions. One string compare per loop iteration.
- Direct paths report too: `ca ssh` attach and `--resume` attach report the listed name up front; a fresh or revived session resolves its relay-listed name in the background (the minted name rides only as an env stamp) and reports it while the session is still running, so a restore prompted by a crash has a working reference. Disconnect re-reports the listing-verified name alongside the printed hint.

Herdr 0.8.0 acknowledges these reports but only persists agent kinds from its built-in integrations — verified against a live server. Reporting the honest label anyway is deliberate: it is the shape herdr's side of the integration can key on, and the reports are inert until then.

# Testing

- `cargo test` (1281 passing; new tests: pane-contract gating via injected env, request framing/shape, and an end-to-end send against a real `UnixListener` including the dead-socket silence).
- `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features` (no new warnings).
- Live against a running herdr 0.8.0 server: the exact request shape is acknowledged over the real socket; storage behavior per agent kind/source verified with a scratch pane (known kinds persist, unknown kinds are acked and dropped — the herdr-side feature this pairs with).

https://claude.ai/code/session_01XEpBNh4vTGb12m1ZV2XLDy

### DIFF
--- a/src/commands/cloud_agent/herdr.rs
+++ b/src/commands/cloud_agent/herdr.rs
@@ -1,0 +1,230 @@
+//! Report the pane's cloud-agent session to Herdr, when this process runs
+//! inside one of its panes.
+//!
+//! Herdr (herdr.dev) is a terminal workspace manager for coding agents: it
+//! records which agent conversation each pane holds and, after a restart,
+//! relaunches the pane back into that conversation. Its integrations for
+//! local agents report over a unix socket the pane's environment names —
+//! `HERDR_ENV=1`, `HERDR_SOCKET_PATH`, `HERDR_PANE_ID` — with one
+//! `pane.report_agent_session` request per change. A `railway code` session
+//! is invisible to that machinery today: the harness runs on the VM, where
+//! the socket does not exist, so the pane records nothing and a restore
+//! lands on a bare shell.
+//!
+//! This module is the local half of the fix: the CLI itself reports the
+//! `--resume` reference (`<agent id>:<durable session name>`) for the session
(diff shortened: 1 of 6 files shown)


itsRems:
Closing: this only pays off with workspace-manager-side support for persisting externally-reported sessions, which isn't planned. #1150 stands alone and is unaffected.

PR #1156, opened 2026-08-31
Confirm generated domain creation

### LINKED ISSUE #1118: railway domain (no subcommand) silently creates a public domain with no confirmation — dangerous default for a command that reads as inspection
I'm an AI coding agent (Claude Code) working through a production deploy. Twice in one session, running `railway domain --service render-service --environment production` to *check* whether a domain existed on a private-networking-only service instead **created** a new public domain, with zero confirmation or warning.

**Why I ran it that way:** the command's own `--help` describes it as multi-purpose — "Add, list, inspect, update, or delete domains for a service" — and takes `--service`/`--environment` flags that read like status-query flags (they match the pattern of `railway status --json --service X --environment Y`, which *is* read-only). Nothing about invoking it bare signals "this will create infrastructure if none exists." The create-by-default behavior is documented, but only inside `--help`'s fine print ("Running without a subcommand preserves the original create behavior"), not in a way that's visible before running it.

**Why it's not just a minor inconvenience here:** the service in question (a headless Chromium PNG renderer) is deliberately designed to be reachable only over Railway's private network plus a shared secret — no public domain, ever, by design. Running what looked like a status check created a real, unauthenticated-until-noticed public entry point into it, in production, with no prompt to confirm.

**Suggested fix:** bare `railway domain` (no subcommand) should not silently create. Either:
- require an explicit `--create` (or similar) flag to create, with bare invocation instead delegating to `list`-like behavior, or
- print a clear "no domain exists — creating one now, pass `--dry-run`/`-y` to control this" line *before* creating, not after.

Related but distinct from #985 (agent-harness-detected consent inferring `railway up` auto-create): this one isn't gated by any agent-detection heuristic — it's the unconditional default for every caller, human or agent, who runs `railway domain` expecting a status check.

Repro:
```
railway domain --service <a service with no existing domain> --environment production
```
Expected: shows there's no domain, or errors asking what to do.
Actual: creates one immediately.

anishfyi:
## Summary

Fixes #1118.

Require confirmation before bare `railway domain` creates a Railway-provided public domain. The prompt defaults to no, and non-interactive use must pass `--yes`. Existing domain inspection and explicit custom-domain creation behavior remain unchanged.

Added parser and confirmation-policy regression coverage, including the non-interactive safety path and `--yes` bypass.

## Validation

- `cargo clippy --all-targets --all-features`
- `cargo test` (1,274 tests)

### DIFF
--- a/src/commands/domain.rs
+++ b/src/commands/domain.rs
@@ -19,7 +19,7 @@ use super::*;
 /// Add, list, inspect, update, or delete domains for a service.
 ///
 /// Running without a subcommand preserves the original create behavior:
-/// - `railway domain` generates a Railway-provided service domain
+/// - `railway domain` confirms before generating a Railway-provided service domain
 /// - `railway domain example.com` creates a custom domain
 #[derive(Parser)]
 #[clap(
@@ -50,6 +50,10 @@ pub struct Args {
     /// Specifying a custom domain will also return the required DNS records
     /// to add to your DNS settings.
     domain: Option<String>,
+
+    /// Skip confirmation when creating a Railway-provided domain
+    #[clap(short = 'y', long = "yes")]
+    yes: bool,
 
     /// Output in JSON format
     #[clap(long, global = true)]
@@ -122,6 +126,7 @@ pub async fn command(args: Args) -> Result<()> {
         environment,
         project,
         domain,
+        yes,
         json,
     } = args;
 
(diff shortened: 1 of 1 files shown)


anishfyi:
The label check requires a maintainer-applied release label. Could you please add `release/patch` if that is the appropriate classification for this fix?

PR #1157, opened 2026-08-31
Fix SSH with explicit project names

### LINKED ISSUE #1104: `railway ssh` requires a link even when project + environment are passed
```
D venv ❯ railway ssh --project "toth-preview-environments" --environment "production" --service "feat/facebook-business-sdk-v25" 'uv run --with polyfactory --script migrations/seed.py'
Project not found. Run `railway link` to connect to a project.
  → Run `railway link` to connect to a project.
```

`railway ssh` should not need a linked project if a `--project` and `--environment` are defined.

anishfyi:
## Summary

Fixes #1104.

Allow `railway ssh --project <name-or-id> --environment <environment> --service <service>` to work outside a linked project directory. Explicit project values now use the existing project ID-or-name resolver before fetching project details.

A regression test covers resolving a project name with no local link and verifies that the resolved project ID is used for the project query.

## Validation

- `cargo clippy --all-targets --all-features`
- `cargo test` (1,274 tests)

### DIFF
--- a/src/commands/ssh/common.rs
+++ b/src/commands/ssh/common.rs
@@ -4,7 +4,9 @@ use reqwest::Client;
 use crate::commands::queries::RailwayProject;
 use crate::config::Configs;
 use crate::controllers::{
-    environment::get_matched_environment, project::get_project, service::get_or_prompt_service,
+    environment::get_matched_environment,
+    project::{get_project, resolve_project_id_or_name},
+    service::get_or_prompt_service,
 };
 
 use super::Args;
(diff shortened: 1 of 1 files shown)


anishfyi:
The label check requires a maintainer-applied release label. Could you please add `release/patch` if that is the appropriate classification for this fix?

anishfyi:
I also covered the explicit `--project` plus `--environment` case when `--service` is omitted. It no longer requires a local link and directly selects the project's sole service. The expanded regression test, full `cargo test`, formatting, and `git diff --check` pass. Normal Clippy completes with the repository's existing unrelated warnings.

PR #1158, opened 2026-08-31
fix(database): ha revert must not delete orphans it cannot attribute

paulocsanz:
#1153 added a resumed sweep so a revert that died mid-delete could not strand members *"still deployed and still billing with no command able to remove them."* That goal is right. The problem is **which** services it deletes.

## The match is environment-wide, and cannot be otherwise

`revert_sweep_targets`'s debris scan:

```rust
cluster_role in (replica, internal, edge)
  && parent_service_id.is_none()
  && !is_deleted && id != root_id && !is_pooler
```

Nothing there ties a candidate to the root being reverted — and nothing can. `templateRevert` clears the survivors' parent links *before* deleting them, so by the time a resumed sweep runs there is no link left to scope by. (`find_root_id` would exclude the very debris the sweep exists for, and `groupId` is explicitly *"not a record of group membership"*.)

## Why that is dangerous, not just imprecise

**The branch is not narrow.** `is_cluster` is just "the root's HA-active variable reads true", so `ha revert` takes the resumed-sweep path for **every ordinary standalone database**, not only one that used to be a cluster. Before #1153 all of those bailed safely with `"... is not an HA cluster."`

**Another cluster can hold services in exactly the matched state.** Most obviously when its own root Service row was deleted: `ServiceInstance.parentServiceId` is `onDelete: SetNull`, which nulls every child's link and leaves `clusterRole` untouched. Those members are still deployed and still billing — indistinguishable from this root's debris.

**Neither guard caught it.** The prompt showed only a count (*"N cluster member(s) from an earlier revert"*), so an operator could not see that N included another cluster's services; the names were printed only inside the deletion loop, after consent. And `--yes` skipped the prompt entirely, so a scripted `railway <engine> ha revert --yes` against an ordinary standalone could delete a different cluster's live members.

Worth noting the scan is now engine-generic, so this reaches MySQL and Redis HA too, not just Postgres.

## The fix

The attribution genuinely cannot be recovered, so this does not pretend to scope the match. It stops treating the match as proof:

- **Every candidate is printed by name** before anything is deleted, with a plain statement that they cannot be attributed to this root.
- **A new `--remove-orphans` opts in, separately from `--yes`.** Deleting services we cannot attribute should not ride on a generic confirmation-skip flag.

The in-cluster path is unchanged — it pairs the same scan with a real membership snapshot and an explicit revert the operator asked for.

The extra flag is a deliberate UX change on a path that previously ran under `--yes`; happy to reshape it if you would rather keep one flag, but the count-only prompt plus `--yes` is the combination that makes this deletable-by-accident.

## Verification

`the_resumed_sweep_matches_another_clusters_orphans_too` pins the hazard: reverting an untouched standalone `db-b` returns another cluster's `db-a-replica` and `db-a-haproxy`, while an intact cluster's members (parent links present) never match. **If a future change makes the scan scopable, that test fails and the gate can come off with it.** The arg-parsing test now asserts `--yes` alone leaves `remove_orphans` false.

`cargo test` 1304 passing, `rustfmt` clean, no new clippy warnings.

Found by an adversarial review of merged changes; the executable PoC is at `.claude/audit/consolidated/proof/cli-1153-revert-sweep/` in the mono repo.

### DIFF
--- a/src/commands/database/ha.rs
+++ b/src/commands/database/ha.rs
@@ -91,6 +91,18 @@ struct RevertArgs {
     #[clap(long, short = 'y')]
     yes: bool,
 
+    /// Also delete role-stamped services in this environment that run this
+    /// engine's cluster images but have no cluster parent left (debris from
+    /// a revert that died mid-sweep).
+    ///
+    /// Separate from --yes on purpose: an earlier revert clears the parent
+    /// links before it deletes, so nothing on these services still ties them
+    /// to the root you named -- image lineage narrows the match to this
+    /// engine, but two clusters of the same engine in one environment remain
+    /// indistinguishable.
+    #[clap(long)]
+    remove_orphans: bool,
+
     /// Commit the config change without triggering deploys (applies on the next deploy)
     #[clap(long)]
     no_deploy: bool,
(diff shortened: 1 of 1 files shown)


paulocsanz:
Rebased onto v5.47.0 (`d2b574c`, replacing `63334b9`). #1159 landed in between and changed this PR's premise, so the rebase is semantic, not just textual — what changed:

- **The scan is now scoped** (to the images the engine's companion publishes, read off the companion record), so the original "the match cannot be otherwise" framing and the `NOTE` comment are gone. The gate itself survives on the narrower residual hazard: image lineage keeps another *engine's* debris out but cannot tell two clusters of the **same engine** apart, and the resumed-sweep branch is still reachable from any ordinary standalone.
- `the_resumed_sweep_matches_another_clusters_orphans_too` was reworked into `the_resumed_sweep_matches_a_same_engine_clusters_orphans_too`: the cross-engine fixture would no longer match (that's #1159's test now), so it pins the same-engine case instead — reverting an untouched standalone `db-b` matches another *Postgres* cluster's orphaned members. Same tripwire contract as before: if the scan ever becomes scopable to the named root, the test fails and the gate can come off with it.
- The upfront named listing now states what the attribution rests on (image lineage, not membership); `--remove-orphans` and the `--yes`-must-not-arm-it parse assertion are unchanged in spirit.

`cargo test` 1311 pass, clippy/fmt clean. If you'd rather drop the gate entirely now that the scan is scoped, that's a one-commit revert of this branch — the tripwire test is the part I'd keep either way.

(Note for anyone with a local checkout at `cli-wt-sweep`: the branch was force-pushed; `git fetch && git reset --hard origin/pcs/ha-revert-sweep-scoped` to sync.)

PR #1162, opened 2026-08-31
fix(database): read suffixed image tags the way the server-side gate does

paulocsanz:
Follow-up to #1143 (same audit as #1160/#1161).

`image_tag_version` required purely numeric tag components; the platform's gate (`extractImageTagVersion` in backboard, regex `^(\d+)(?:\.(\d+))?`) parses the **leading** `major.minor` run and ignores the variant suffix. Since these checks pre-flight that gate, the stricter parse turned the pre-flight into a false blocker:

- `redis:8.2-alpine` read as "declares only a major" and was refused for a `pinToMinorVersion` conversion the backend accepts (and pins to 8.2).
- `redis:7-bookworm` read as carrying no version at all — refused with "no version can be read" and a spurious supported-majors complaint, instead of the accurate "declares only a major" with the retag remedy. Bookworm-suffixed tags are what the frozen pre-rebase redis sub-fleet still runs, and conversion is its patch path.

Now parsed identically to the platform: leading digits are the major, a following `.digits` run is the minor, anything after is ignored, and a non-digit lead is no version at all.

## Verification

`cargo test` — 1304 pass, with new coverage pinning `8.2-alpine` / `7-bookworm` / `7.2.4-debian` at both layers: the parser itself, and the conversion pre-flight's verdicts (accepted; refused with the retag remedy). `cargo clippy` / `cargo fmt` clean.

### DIFF
--- a/src/controllers/adoption_eligibility.rs
+++ b/src/controllers/adoption_eligibility.rs
@@ -448,6 +448,38 @@ mod tests {
     }
 
     #[test]
+    fn suffixed_tags_are_read_the_way_the_server_side_gate_reads_them() {
+        // The platform's gate parses the LEADING major.minor of a tag and
+        // ignores the variant suffix, so the pre-flight must too -- refusing
+        // `redis:8.2-alpine` for "declaring only a major" blocked a
+        // conversion the backend accepts (and pins to 8.2).
+        let rules = rules_from_template(&json!({
+            "services": {
+                "root": {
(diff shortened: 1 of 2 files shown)


paulocsanz:
Closing: #1159 merged the same fix (leading major[.minor] parse mirroring the server's extractImageTagVersion) and shipped in v5.47.0 — a rebase leaves this PR empty.

PR #1165, opened 2026-09-02
fix(database): 10-minute wait for cluster stage/commit workflows, with a timeout message that says what is left to do

paulocsanz:
## Problem

`wait_for_workflow` gives every workflow the same fixed ~2-minute budget (120 one-second polls). `controllers/template_apply.rs` uses it for the stage workflows of `templateDeployV2` / `templateRevert` and for the `commitPatchToEnvironment` workflow that applies the committed patch. Those workflows create and delete every member service of a cluster, so their duration scales with the topology asked for and with Temporal's queue latency.

Live evidence from the postgres-cli e2e harness (`paulocsanz/test-postgres-cli`, cron on Railway): `postgres ha convert --replicas 2 --coordinators 3 --edge 2` finished inside the budget at 2026-09-01 17:12Z and exceeded it at 23:14Z the same day, exiting 1 with

```
Failed to convert to HA

Caused by:
    workflow timed out
```

for a conversion that was committed and applying. A retry was then refused by backboard's in-flight guard (mono #37352, `A cluster change for this service is still being applied`), so the user was told the conversion failed, saw the retry refused, and was given nothing to do.

## Change

- The three waits share a 10-minute budget (`CLUSTER_WORKFLOW_WAIT_ATTEMPTS = 600`) via `wait_for_workflow_up_to`, the same mechanism `pitr backup restore` already uses for its data-dependent copy.
- Each timeout says what state the environment is in and what, if anything, is left to do: after a stage timeout, the changes appear as staged changes to apply or discard, and a retry is refused until then; after a commit timeout, nothing — the change keeps applying.
- Behaviour on success is unchanged. `cargo check` clean; `controllers::template_apply` tests 20/20.

## Harness side

The e2e harness stops treating this timeout as a failed conversion independently (paulocsanz/test-postgres-cli, same day), so the cron is green before this ships; this PR fixes what a real user sees.

### DIFF
--- a/src/controllers/template_apply.rs
+++ b/src/controllers/template_apply.rs
@@ -265,8 +265,13 @@ pub async fn apply_composable_template(
     .with_context(|| format!("Failed to deploy template \"{}\"", params.template_code))?;
 
     if let Some(workflow_id) = response.template_deploy_v2.workflow_id.clone() {
-        crate::controllers::workflow::wait_for_workflow(&ctx.client, &ctx.configs, workflow_id)
-            .await?;
+        wait_for_cluster_workflow(ctx, workflow_id, |minutes| {
+            format!(
+                "Template \"{}\" is still being staged server-side after {minutes} minutes. When staging finishes it appears as staged changes on the environment -- apply them to finish, or discard them to start over. Until then a retry of this command is refused as a cluster change still in flight.",
+                params.template_code
+            )
+        })
+        .await?;
     }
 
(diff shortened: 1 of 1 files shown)


paulocsanz:
Closing by Paulo's ruling: the CLI keeps the default workflow wait and fails with it when the apply workflow outlives it. Callers that need to outlive that wait own the waiting (the e2e harness does, paulocsanz/test-postgres-cli#13).

PR #1176, opened 2026-09-08
experiment: herdr integration for cloud-agents

lawrencegripper:
## What changed

- `railway ca herdr {install,new,agents,sync,bootstrap}`: cloud agents appear in herdr 0.9 as saved SSH machines. `install` writes and links a herdr plugin manifest whose every command calls back into these verbs, so the plugin ships with the CLI and has no code of its own.
- `bootstrap` provisions the VM in app mode (the same `code::prepare` as `ca desktop`), installs herdr's Claude and Codex integrations, sets `[terminal] shell_mode/default_shell` and `[experimental] pane_history`, adds a `~/.profile` block sourcing the carried token, and creates an `/app` workspace.
- The relay host key is seeded into `~/.ssh/known_hosts` from the CLI's `~/.railway/known_hosts_relay`.

## Why

herdr 0.9 runs one server per machine, and a cloud agent VM is exactly that: herdr owns the panes and detects the agent natively, and after a sleep its own restore brings the layout and `claude --resume` back. herdr cannot create, wake, sleep or delete the VM, which is what these verbs do. Relay durable sessions are not used.

The ssh target is `ssh://agent%3A<env>%3A<id>@ssh.railway.com`: herdr rejects a literal colon in the user part as a password, OpenSSH decodes the percent form. herdr's background reconnects run with `StrictHostKeyChecking=yes`, so the relay key has to be in `~/.ssh/known_hosts`. Pane shells on the VM fall back to `/bin/sh` because the server inherits no `SHELL`, hence `default_shell`.

## Notes

- Experiment. Proven on one agent (connect, sleep, wake, resume) against an isolated herdr session; `new` against a real create not yet run.
- `lifecycle::place_names` and `code::HARNESS_PATH` made `pub(crate)`.
- Design notes live in `~/claude-brain/projects/2026-09-08-herdr-railway-ca-plugin`.

https://claude.ai/code/session_01Vvwnaf9R6RjfPTrRo7MyDu


codyde:
Opened #1187 targeting this branch with the pressure-tested reliability fixes: account/backend-scoped and atomic state updates, stale-plan checks around relay waits, server-authoritative wake/sleep requests, reconnect refetches plus periodic reconciliation, SSH identity handoff, and empty-picker/bootstrap retry fixes.

All 1,418 tests pass with injected Railway token overrides cleared from the test process; formatting and Clippy checks pass (existing lint warnings). The contribution includes the reproductions and the remaining integration limits: Herdr 0.9 already retries ordinary network failures, but its machine-list API cannot identify live Attention state, and live per-harness create/sleep/wake restoration still needs end-to-end validation.

PR #1181, opened 2026-09-09
fix(config): fail apply when the change-set workflow fails

### LINKED ISSUE #1170: config apply exits 0 and prints nothing when the change-set workflow fails (e.g. duplicate database name)
## Summary

`railway config apply` (both `--plan <pinned>` and the live `--yes --confirm-destructive` form) returned exit code 0 and printed nothing beyond the header while the server-side change set had failed with a diagnostic. Nothing was applied; the environment `configEtag` was unchanged. A CI pipeline built on the documented pattern (`plan --out` on PR, `apply --plan` on merge) therefore reported success for a merge that changed nothing.

## Environment

- `@railway/cli` 5.49.2 (also observed with `railwayapp/config@v1`, which installs `latest`)
- Auth: project token scoped to the target environment (`RAILWAY_TOKEN`)
- GitHub Actions `ubuntu-latest`, and reproduced by re-running the same job

## What happened

Plan for the target environment listed 15–19 changes, including `+ Create database Redis Exporter` (an image-based service that Railway classifies as `database/engine=redis`). The same service already existed in the project — created earlier by `railway config apply` in a **different** environment — with no service instance in the target environment.

- `railway config apply --plan railway-plan.json --yes --confirm-destructive` → exit 0, **no stdout**. (`railwayapp/config` then prints "Applied pinned Railway configuration.")
- `railway config apply --yes --confirm-destructive` → exit 0, stdout only:
  ```
  Railway configuration
  Using .railway/railway.ts
  Project <project>
  Environment production
  ```
  No `Plan:` line, no "Applied", no error.
- Immediately after, `railway config plan` showed the same changes; `configEtag` identical; no events in the environment.

The actual failure was only visible through the API, on the `environmentApplyChangeSet` workflow:

```json
{
  "id": "iac-change-set/<environmentId>/<hash>",
  "status": "failed",
  "changes": [],
  "diagnostics": [{
    "message": "A service named \"Redis Exporter\" already exists in this project but not in this environment. Databases managed by Railway config need a project-unique name. Rename \"Redis Exporter\" in your config, or rename/delete the existing service."
  }]
}
```

## Expected

- Non-zero exit code when the change-set workflow ends in `failed`.
- The `diagnostics[].message` printed to stderr.
- Ideally `plan` would surface the same conflict up front, since it is knowable before apply.

## Related question

Is there a supported IaC path to create the service instance of an *existing* project-level service in a second environment? We ended up using the dashboard's **Sync environments** to create the instance, after which `plan` converged without any apply. If that is the intended workflow, a hint in the diagnostic (or in the IaC docs under "environments") would save the next person a few hours.

## Workaround for others hitting this

Re-plan after every apply in CI (`railway config plan --detailed-exit-code`) and fail the job on exit 2. That is what caught this for us.

siddiqueirshad:
## Summary
- Fixes #1170: `railway config apply` treated a failed change-set workflow (`status: failed`, often with empty `changes` and the real explanation only in `diagnostics`) as success, exiting 0 and printing only the header.
- Reject failed apply payloads in `apply_change_set` so both live apply and `apply --plan` return a non-zero exit with the server diagnostics.
- Defense-in-depth in the config runner for legacy TS-runner responses, plus clearer apply-failure printing when changes are empty.

## Root cause
`environmentApplyChangeSet` / `environmentChangeSetApply` can finish as `failed` with `changes: []` and diagnostics such as a duplicate database name. The CLI stored that payload in `apply_result`, left `ok: true`, printed nothing when `changes` was empty, and exited 0 — so CI (including `railwayapp/config`) reported a successful apply that changed nothing.

## Test plan
- [x] `cargo test ensure_apply_succeeded`
- [x] `cargo test apply_diagnostic`
- [x] `cargo test reject_failed_apply_result_surfaces`
- [x] `cargo test iac::` (74 passed)
- [ ] Repro from #1170: apply a plan that creates a database service whose name already exists project-wide; expect non-zero exit and the diagnostic message on stderr

### DIFF
--- a/src/commands/config/runner.rs
+++ b/src/commands/config/runner.rs
@@ -282,6 +282,7 @@ pub(super) async fn run_command(args: Args) -> Result<()> {
         if !output.ok {
             bail!(runner_diagnostics_message(&output));
         }
+        reject_failed_apply_result(output.apply_result.as_ref())?;
         maybe_detailed_exit(&args, command, &output);
         return Ok(());
     }
@@ -290,6 +291,7 @@ pub(super) async fn run_command(args: Args) -> Result<()> {
     if !output.ok {
         bail!(runner_diagnostics_message(&output));
     }
+    reject_failed_apply_result(output.apply_result.as_ref())?;
 
     maybe_detailed_exit(&args, command, &output);
 
@@ -745,6 +747,26 @@ fn runner_diagnostics_message(response: &RunnerResponse) -> String {
     )
 }
 
+fn reject_failed_apply_result(apply_result: Option<&ChangeSetApplyResult>) -> Result<()> {
+    let Some(apply_result) = apply_result else {
+        return Ok(());
+    };
+    if !crate::iac::apply_status_is_failure(&apply_result.status) {
+        return Ok(());
+    }
+    let messages = crate::iac::apply_diagnostic_messages(&apply_result.diagnostics);
+    if messages.is_empty() {
+        bail!(
+            "Railway configuration apply failed (status: {}).",
+            apply_result.status
+        );
+    }
+    bail!(
+        "Railway configuration apply failed:\n{}",
+        messages.join("\n")
+    )
+}
+
 pub(super) fn print_response_with_options(response: &RunnerResponse, verbose: bool) {
     print_response_with_options_and_next(response, verbose, true);
 }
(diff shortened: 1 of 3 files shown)


siddiqueirshad:
Hi @brody192, when you get a chance, could you please take a look at this PR? Thanks!

PR #1182, opened 2026-09-09
fix(release): unblock i686 Windows GNU builds

codyde:
## Summary

- Pin the `i686-pc-windows-gnu` cross image to an upstream Ubuntu 24.04 / GCC 13 image by digest.
- Use MinGW's import libraries consistently via `WINAPI_NO_BUNDLED_LIBRARIES=1` in `Cross.toml`.
- Add a full release-mode cross-build to CI using the same checkout, Rust toolchain, cache, and `actions-rs/cargo@v1` structure as the existing checks. The build action and arguments match `release.yml`.

- Serialize the Unix upgrade tests’ executable fixture writes and child execution to avoid intermittent Linux `ETXTBSY` failures.

## Why

The releases containing #1173 (`v5.50.0` and `v5.50.1`) are stuck as drafts because the 32-bit Windows GNU build fails with:

```text
multiple definition of `_imp__InterlockedCompareExchange@12'
```

Bundled SQLite stores the address of this function in its Windows syscall table, exposing a collision between MinGW's `libkernel32.a` and winapi's bundled `libwinapi_kernel32.a`. The successful preceding release used the same Rust version, cross version, and container digest; the release workflow was unchanged.

The first CI attempt confirmed the library setting removes that collision but exposed a second issue: cross 0.2.5's default GCC 7.5 / MinGW image lacks `libsynchronization.a`. The pinned modern upstream image provides that library and the other import libraries needed by the full CLI. Both CI and the existing release workflow read the same target-specific `Cross.toml` configuration.

Image source: cross-rs/cross revision `8c1a8aa4b661711f4b7b6ac07c2e8929ce2f7d27`, built 2026-09-04.

The other 11 release targets succeeded, but the failed matrix job prevents crates.io, GitHub Release, and npm publication. Normal native CI does not build this target.

Failed releases:
- https://github.com/railwayapp/cli/actions/runs/34293256109/job/102284295599
- https://github.com/railwayapp/cli/actions/runs/34308453562/job/102330015059

## Validation

- Reproduced the exact original duplicate-symbol error in a minimal C program using the release's old cross container; the same code links using only MinGW's libraries. [Reproduction details](https://github.com/railwayapp/cli/pull/1182#issuecomment-5603818197).
- Verified the pinned image supplies `libsynchronization.a` and links the reproduction with all ten libraries requested by winapi.
- `git diff --check` passed.
- `actionlint` passed for the changed CI workflow with existing diagnostic categories excluded (legacy cargo action and undefined `docker-prep` output).
- All eight PR checks passed on `8fe0abb`: full Windows GNU release build, Ubuntu/macOS/Windows tests, cargo check, lints, POSIX installer, and release label. [Successful CI run](https://github.com/railwayapp/cli/actions/runs/34367819127).
- `cargo fmt --all -- --check` passed after the fixture change.

## Release

The `release/patch` label will trigger a new patch release when merged, including the already-merged OpenCode features and this build fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,30 @@ jobs:
         with:
           command: check
 
+  windows-gnu-release:
+    name: Release build (i686-pc-windows-gnu)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: i686-pc-windows-gnu
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: i686-pc-windows-gnu-release
+
+      - name: Install cross
+        run: cargo install cross --version 0.2.5 --locked
+
+      # Match release.yml's cross target and release profile: cargo check and
+      # native Windows tests do not catch cross-target linker failures.
+      - name: Build release binary
+        run: cross build --release --locked --target i686-pc-windows-gnu
+
   lints:
     name: Lints
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,16 +39,15 @@ jobs:
           targets: i686-pc-windows-gnu
 
       - uses: Swatinem/rust-cache@v2
-        with:
-          key: i686-pc-windows-gnu-release
-
-      - name: Install cross
-        run: cargo install cross --version 0.2.5 --locked
 
       # Match release.yml's cross target and release profile: cargo check and
       # native Windows tests do not catch cross-target linker failures.
-      - name: Build release binary
-        run: cross build --release --locked --target i686-pc-windows-gnu
+      - name: Run cargo build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release --locked --target i686-pc-windows-gnu
+          use-cross: true
 
   lints:
     name: Lints
--- a/Cross.toml
+++ b/Cross.toml
@@ -1,0 +1,5 @@
+[target.i686-pc-windows-gnu.env]
+# Bundled SQLite links MinGW's kernel32 alongside winapi's bundled import
+# libraries, which both define _imp__InterlockedCompareExchange@12 on i686.
+# Use the toolchain's import libraries consistently inside the cross container.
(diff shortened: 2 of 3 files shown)


codyde:
Additional root-cause validation:

The last successful `v5.49.6` build and the failed `v5.50.0` build used the same Rust 1.98.1, cross 0.2.5, and cross-container digest (`sha256:c4b541d5baa76a1f13390850ad65ac5daa811c701842802b655edb3eb832c575`). The release workflow was unchanged. #1173 added bundled SQLite.

SQLite's `sqlite3.c` stores the address of `InterlockedCompareExchange` in its Windows syscall table. I reproduced the exact linker failure with this minimal C program in the same cross container:

```c
#include <windows.h>
static LONG (WINAPI *volatile compare_exchange)(LONG volatile *, LONG, LONG) =
    InterlockedCompareExchange;
int main(void) {
    volatile LONG value = 0;
    return compare_exchange(&value, 1, 0);
}
```

Results with `i686-w64-mingw32-gcc -Os`:

| Case | Result |
| --- | --- |
| Function pointer, `-lkernel32 -lwinapi_kernel32` (winapi 0.4.0 archive) | Fails with the exact `_imp__InterlockedCompareExchange@12` duplicate-symbol error and archive members seen in both releases |
| Same function pointer, only `-lkernel32` | Links successfully |
| Ordinary direct `InterlockedCompareExchange` call, both library sets | Links successfully |

This isolates the failure mechanism and explains why it was exposed by adding SQLite despite previous successful builds. `WINAPI_NO_BUNDLED_LIBRARIES=1` is winapi's supported option to select the toolchain libraries. Full CLI cross-build validation is tracked by the new CI job.

codyde:
The full `i686-pc-windows-gnu` release build now passes with the pinned MinGW image and toolchain import libraries:
https://github.com/railwayapp/cli/actions/runs/34366648604/job/102516975033

The remaining red check was an existing upgrade integration test failing to spawn its temporary binary with Linux `ETXTBSY` (`Text file busy`). All 1,424 unit tests passed; the test matrix then cancelled macOS and Windows.

Commit `8fe0abb` serializes executable fixture writes and child execution within `tests/upgrade_flow.rs`. Concurrent process launches can inherit another test's writable copy handle until exec, preventing that copied binary from being launched. Keeping the fixture operations under one mutex removes that overlap while preserving the test assertions.

Formatting and diff checks pass. The full CI suite is running again on the updated commit.

PR #1186, opened 2026-09-10
fix(ca): preserve agent identity across uncertain VM observations

codyde:
## What this fixes

When cloud-agent status is delayed or unavailable, the CLI can create a second VM instead of reconnecting to your existing one. It can also skip a wake or sleep request because an old status says the agent is already awake or asleep.

This PR makes `railway ca` keep the agent you selected and handle wake/sleep requests more reliably. The agent-selection fixes also apply to `railway code`.

## What changes

- **Keep your existing agent.** If the selected or remembered agent is missing or reports a problem, show an error instead of silently creating a replacement or choosing another agent.
- **Always send wake/sleep requests.** Let the server decide whether anything needs to happen, rather than skipping the request based on old status.
- **Make TUI progress trustworthy.** Show reported failures promptly. Ignore late replies from older status checks so they cannot incorrectly finish a newer action.
- **Handle multiple pending actions.** Poll each affected environment fairly and respect the API's retry delay when rate-limited.
- **Say what actually happened.** Sleep commands report “Sleep requested,” because accepting the request does not mean the VM has finished sleeping.

## Testing

- **1,445 unit tests and 18 integration tests passed** on Linux.
- Added tests for out-of-order replies, concurrent wake/sleep actions, missing agents, and partial failures in `ca sleep --all`.
- Command-level tests run the compiled CLI against a local API server and check its requests, output, and exit codes.
- Formatting and Clippy passed; existing warnings remain.

### DIFF
--- a/src/commands/cloud_agent/lifecycle.rs
+++ b/src/commands/cloud_agent/lifecycle.rs
@@ -351,27 +351,18 @@ pub async fn create(args: CreateArgs) -> Result<()> {
 pub async fn wake(args: WakeArgs) -> Result<()> {
     let mut configs = Configs::new()?;
     let client = GQLClient::new_authorized(&configs)?;
-    let (configs, client) = (&mut configs, &client);
+    wake_with(&mut configs, &client, args).await
+}
+
+async fn wake_with(configs: &mut Configs, client: &reqwest::Client, args: WakeArgs) -> Result<()> {
     let (project, environment) = (args.target.project.clone(), args.target.environment.clone());
     let scoped = scope(configs, client, project, environment).await?;
     let (agent, _) = ca::resolve(configs, client, args.agent.as_deref(), scoped.as_deref()).await?;
     let backboard = configs.get_backboard();
 
-    match agent.status {
-        ca::Status::Running => {
-            println!("Agent {} is already awake.", agent.name.cyan());
-            ca::remember(configs, &agent)?;
-            return Ok(());
-        }
-        ca::Status::Sleeping => ca::wake(client, &backboard, &agent.id).await?,
-        // Something else is already booting it; a second wake would be noise.
-        ca::Status::Starting => {}
-        _ => bail!(
-            "Agent {} is {} and cannot be woken.",
-            agent.name,
-            agent.status.label()
-        ),
-    }
+    // The inventory is an observation. Let the mutation's live state check
+    // decide whether waking is necessary, including when we last saw RUNNING.
+    ca::wake(client, &backboard, &agent.id).await?;
     ca::remember(configs, &agent)?;
 
     if args.no_wait {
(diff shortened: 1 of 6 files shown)


codyde:
## Pressure-test follow-up

Added adversarial coverage and fixed the issues it exposed:

| Failure scenario | Finding and fix |
| --- | --- |
| A pre-mutation snapshot arrives after wake/sleep acknowledgement | Reproduced premature watch completion. Both query paths now compare request-start times against a per-environment floor advanced by successful mutations and applied snapshots. |
| Overlapping scoped/account-wide responses arrive out of order | Use request-start ordering on both paths rather than response-arrival time. Tested all 6 arrival orders across all 8 query-source combinations (48 scenarios). |
| Two agents have pending operations in different environments | Reproduced polling only one environment indefinitely. Watch polling now rotates through distinct environments. |
| The API returns Retry-After during a watch | Reproduced continued polling. Watch polls now honor the pause while still expiring overdue watches. |
| A remembered agent is missing from inventory | Reproduced automatic selection falling through to another agent or empty inventory. Bare commands now preserve the missing target and require an explicit selection or creation. Tested environment scoping and explicit overrides. |
| Tree order changes while a scoped fetch is in flight | Responses now carry the environment identity and cannot be filed into a different environment's row. |
| A manual account refresh discovers a failed operation | Its generic “Up to date” completion message no longer overwrites the operation failure. |

The first four targeted regressions (pre-mutation response, polling starvation, Retry-After, missing remembered selection) were run before the corrections: **4 failed**. All have permanent regression tests.

Added black-box tests of the compiled CLI against a local GraphQL server, with an isolated HOME and fake authentication:

- Single sleep across every known status and a future enum value, including best-effort disk-flush failure.
- Bulk sleep success and partial failure: all agents attempted, correct accepted-request count, nonzero exit on partial failure.
- Single sleep refusal: no success message.
- Wake with an initially RUNNING inventory that subsequently reports CRASHED; `--no-wait` verifies zero polling.
- `ca ssh` with a missing remembered target: error before replacement creation.

### Validation

- **1,445 unit tests passed**, including the new timing, selection and concurrency regressions.
- **18 integration tests passed** across all four integration targets, including 5 new cloud-agent command tests.
- `cargo clippy --locked --all-targets --all-features`: passed with the same existing warnings; none in the new code.
- `cargo fmt --all -- --check` and `git diff --check`: passed.

The broad run initially exposed authentication-test contamination from the VM's inherited API token. Removing `RAILWAY_API_TOKEN` and `RAILWAY_TOKEN` from the test process environment made those tests pass. A new bulk-output assertion was also corrected to match the CLI's existing em-dash separator before all integration targets passed. No application behavior was changed to accommodate those test-environment/assertion issues.

Tests ran on Linux with Rust 1.98.1, one compilation job, debug information/incremental compilation disabled, and LLD for the memory-limited development VM.

### Boundary of these fixes

These tests exercise CLI behavior and real HTTP/GraphQL serialization, not live FactoryVM lifecycle transitions. Local request timestamps prevent local response-ordering bugs; they cannot prove that a backend projection reflects the latest VM generation. A failed wake that returns to SLEEPING and the distinction between VM failure and unavailable observation still need the richer backend contract discussed in the original review.

